PR #40, opened 2023-09-25
fix: use https path for libpg_query submodule

imor:
I was unable to checkout the submodule without the new https path.


psteinroe (approved):
Thanks!

PR #46, opened 2023-11-11
add test for parsing BEGIN, ROLLBACK, COMMIT statements

### LINKED ISSUE #36: add tests for (and maybe fix) transaction statements `BEGIN;`...`COMMIT;`
# Bug report

## Describe the bug

I am not sure yet whether this is a bug, but we should add tests for transaction statements to see how they are parsed by lipg_query. They should be parsed as a single statement. We may have to adapt the `source_parser` regex for that.

## Expected behavior

Transactions are beneath a single ast node, and a nested cst node.

pert5432:

Added test

Transaction statement parsing was not tested

## What is the new behavior?

Transaction statement parsing is now tested

closes #36

### DIFF
--- a/crates/parser/tests/data/statements/valid/0037.sql
+++ b/crates/parser/tests/data/statements/valid/0037.sql
@@ -1,0 +1,7 @@
+BEGIN;
+
+SELECT 1;
+
+ROLLBACK;
+
+COMMIT;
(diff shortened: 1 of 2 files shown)


psteinroe:
hey @pert5432, thanks for the contribution! Sorry it took so long. We finally finished the parser implementation. Would you mind re-trying the pr? we dropped the snapshot tests, so if you could remove the snapshot and re-try the test that would be great. Thanks a lot. 

Also, check out #51 if you have continued interest to contribute.

pert5432:
Yo, I'll re-do it, no problem.

Issue #51, opened 2023-12-05
help finish the parser


psteinroe:
We recently finished the implementation of the core parser algorithm. Details can be found in [this blog post](https://supabase.com/blog/postgres-language-server-implementing-parser).

While a lot of work has been automated with procedural macros, we still need to provide the keywords for a node [in `get_node_properties`](https://github.com/supabase/postgres_lsp/blob/c8f1708241717bff1fbb8762aed422b6806675fe/crates/codegen/src/get_node_properties.rs#L146) manually. For example, a `SelectStmt` node has the `select` keyword as a property, and if there is a `from_clause`, a `from` keyword. This is implemented as

```rust
fn custom_handlers(node: &Node) -> TokenStream {
    match node.name.as_str() {
	"SelectStmt" => quote! {
	        tokens.push(TokenProperty::from(Token::Select));
	        if n.from_clause.len() > 0 {
	               tokens.push(TokenProperty::from(Token::From));
	        }
	        ...
	},
        ...
}
```

## Requirements

Setup a rust development environment:

1. Install [rustup](https://www.rust-lang.org/tools/install)
2. Install nightly rust via `rustup-init`
```sh
rustup-init --default-toolchain nightly -y
```
3. Fork the repository

## Run your first test

To make it as easy as possible for contributors to get into it, we prepared a simple test suite to quickly assert the tokens for a node.

1. open the file [`crates/parser/src/codegen.rs`](https://github.com/supabase/postgres_lsp/blob/main/crates/parser/src/codegen.rs). you will find a bunch of tests there, e.g. `test_simple_select`. In each test, the `test_get_node_properties` helper is called . The parameters are a valid sql input `str`, the kind of the node we want to test for, and a `Vec<TokenProperty>` with the properties we expect.
5. in a terminal, `cd` into `crates/parser`
6. now, run a test with debugging enabled `RUST_LOG=DEBUG cargo test test_simple_select`. you will see both the entire graph, as well as the node you are testing against within the logs. The test should pass, since we are expecting only the `Select` keyword.

## Become familiar with the codebase

A lot of logic is generated using macros, which makes it harder to find the definition of struct, enums and functions. Most important in this case are `SyntaxKind` and `TokenProperty`.

[...]

PR #61, opened 2023-12-10
add support for parsing "create domain" statement

cvng:

Feature: try to add support for "create domain" statement as requested in #51

Parser panics:

```sh
could not find node for token Token { kind: Create, text: "CREATE", span: 0..6, token_type: ReservedKeyword } at depth 1
```

## What is the new behavior?

Parser returns:

```sh
  CreateDomainStmt@0..37
    Create@0..6 "CREATE"
    Whitespace@6..7 " "
    DomainP@7..13 "DOMAIN"
    Whitespace@13..14 " "
    Ident@14..28 "us_postal_code"
    Whitespace@28..29 " "
    As@29..31 "AS"
    Whitespace@31..32 " "
    TypeName@32..36
      TextP@32..36 "TEXT"
    Ascii59@36..37 ";"
```

## Additional context

Not sure of what exactly should go in the `CreateDomainStmt` branch, but I have tried with these variations and it worked:

```sql
CREATE DOMAIN name [ AS ] data_type [ constraint [ ... ] ]
```

### DIFF
--- a/crates/codegen/src/get_node_properties.rs
+++ b/crates/codegen/src/get_node_properties.rs
@@ -535,9 +535,6 @@ fn custom_handlers(node: &Node) -> TokenStream {
             if n.type_name.is_some() {
                 tokens.push(TokenProperty::from(Token::As));
             }
-            if n.constraints.len() > 0 {
-                tokens.push(TokenProperty::from(Token::Check));
-            }
         },
         _ => quote! {},
     }
--- a/crates/codegen/src/get_node_properties.rs
+++ b/crates/codegen/src/get_node_properties.rs
@@ -529,6 +529,16 @@ fn custom_handlers(node: &Node) -> TokenStream {
         "TypeCast" => quote! {
             tokens.push(TokenProperty::from(Token::Typecast));
         },
+        "CreateDomainStmt" => quote! {
+            tokens.push(TokenProperty::from(Token::Create));
+            tokens.push(TokenProperty::from(Token::DomainP));
+            if n.type_name.is_some() {
+                tokens.push(TokenProperty::from(Token::As));
+            }
+            if n.constraints.len() > 0 {
+                tokens.push(TokenProperty::from(Token::Check));
+            }
+        },
         _ => quote! {},
     }
 }
--- a/crates/parser/src/codegen.rs
+++ b/crates/parser/src/codegen.rs
@@ -91,4 +91,19 @@ mod tests {
             ],
         )
     }
+
+    #[test]
+    fn test_create_domain() {
+        test_get_node_properties(
+            "create domain us_postal_code as text check (value is not null);",
+            SyntaxKind::CreateDomainStmt,
+            vec![
+                TokenProperty::from(SyntaxKind::Create),
+                TokenProperty::from(SyntaxKind::DomainP),
+                TokenProperty::from(SyntaxKind::As),
+                TokenProperty::from(SyntaxKind::Check),
+                TokenProperty::from("us_postal_code".to_string()),
+            ],
+        )
+    }
 }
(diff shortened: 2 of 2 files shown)


crates/codegen/src/get_node_properties.rs
```diff
@@ -529,6 +529,16 @@ fn custom_handlers(node: &Node) -> TokenStream {
...
         },
+        "CreateDomainStmt" => quote! {
+            tokens.push(TokenProperty::from(Token::Create));
+            tokens.push(TokenProperty::from(Token::DomainP));
+            if n.type_name.is_some() {
+                tokens.push(TokenProperty::from(Token::As));
+            }
+            if n.constraints.len() > 0 {
+                tokens.push(TokenProperty::from(Token::Check));
+            }
```
psteinroe, line 540: `check` should be part of the `Constraint` node, right? its currently implement as

```rust
        "Constraint" => quote! {
            match n.contype {
                // ConstrNotnull
                2 => {
                    tokens.push(TokenProperty::from(Token::Not));
                    tokens.push(TokenProperty::from(Token::NullP));
                },
                // ConstrDefault
                3 => tokens.push(TokenProperty::from(Token::Default)),
                // ConstrCheck
                6 => tokens.push(TokenProperty::from(Token::Check)),
                // ConstrPrimary
                7 => {
                    tokens.push(TokenProperty::from(Token::Primary));
                    tokens.push(TokenProperty::from(Token::Key));
                },
                // ConstrForeign
                10 => tokens.push(TokenProperty::from(Token::References)),
                _ => panic!("Unknown Constraint {:#?}", n.contype),
            }
        },
```

similarly, `As` should be part of the `TypeName` node.

psteinroe (changes requested):
thanks for the contribution! can you check the node output with debug logs enabled and ensure that the properties match their nodes? 

also, what would help you in figuring this out faster? I think we can improve here.

cvng:
Got it, I have removed the extra token, thanks!

For the `As` token (which is optional), it has no effect under the `TypeName` node, so I had to keep it under the `CreateDomainStmt` node or the parser still panics with:

```sh
could not find node for token Token { kind: As, text: "AS", span: 29..31, token_type: ReservedKeyword } at depth 2
```

As a side-note: `TypeName` node starts at `location: Some(32)` so maybe this is why?
```sh
1: Node {
    kind: TypeName,
    depth: 1,
    properties: [
        TokenProperty {
            value: Some(
                "text",
            ),
            kind: None,
        },
    ],
    location: Some(
        32,
    ),
},
```

crates/parser/src/codegen.rs
```diff
@@ -87,4 +87,18 @@ mod tests {
...
+    #[test]
+    fn test_create_domain() {
+        test_get_node_properties(
+            "create domain us_postal_code as text check (value is not null);",
+            SyntaxKind::CreateDomainStmt,
+            vec![
+                TokenProperty::from(SyntaxKind::Create),
+                TokenProperty::from(SyntaxKind::DomainP),
+                TokenProperty::from(SyntaxKind::As),
+                TokenProperty::from("us_postal_code".to_string()),
```
cvng, line 100: Here, for the range `0..32` I would expect this order:

```rust
TokenProperty::from(SyntaxKind::Create),
TokenProperty::from(SyntaxKind::DomainP),
TokenProperty::from("us_postal_code".to_string()),
TokenProperty::from(SyntaxKind::As),
```

psteinroe: ahh sorry - the order is not relevant. I fixed the test suite in #62 to ignore the order when comparing the vectors. it should pass after you merge main into you branch. thanks!

psteinroe (approved):
lgtm! once you merge main tests should pass too :)

PR #63, opened 2023-12-12
feat: support "create schema"

cvng:
## What kind of change does this PR introduce?

In the continuation of #61 & #51. Ultimately, being able to parse most of the DDL statements would be nice

## What is the current behavior?

Parser panics:

```sh
could not find node for token Token { kind: Create, text: "CREATE", span: 0..6, token_type: ReservedKeyword } at depth 1
```

## What is the new behavior?

Parser returns:

```rust
        CreateSchemaStmt(
            CreateSchemaStmt {
                schemaname: "test",
                authrole: Some(
                    RoleSpec {
                        roletype: RolespecCstring,
                        rolename: "joe",
                        location: 47,
                    },
                ),
                schema_elts: [],
                if_not_exists: true,
            },
        ),
```

## Additional context

Add any other context or screenshots.

PR #65, opened 2023-12-12
feat: support "create view" with relpersistence

cvng:

Follow-up for #63, #51

Parser panics

## What is the new behavior?

Parser returns:

```rust
        ViewStmt(
            ViewStmt {
                view: Some(
                    RangeVar {
                        catalogname: "",
                        schemaname: "",
                        relname: "comedies",
                        inh: true,
                        relpersistence: "t",
                        alias: None,
                        location: 33,
                    },
                ),
                aliases: [],
                query: Some(
                    Node {
                        node: Some(
                            SelectStmt(
                                SelectStmt {
                                    distinct_clause: [],
                                    into_clause: None,
                                    target_list: [
                                        Node {
                                            node: Some(
                                                ResTarget(
                                                    ResTarget {
                                                        name: "",
                                                        indirection: [],
                                                        val: Some(
                                                            Node {
                                                                node: Some(
                                                                    ColumnRef(
                                                                        ColumnRef {
                                                                            fields: [
                                                                                Node {
                                                                                    node: Some(
                                                                                        AStar(
                                                                                            AStar,
                                                                                        ),
                                                                                    ),
                                                                                },
                                                                            ],
                                                                            location: 52,
                                                                        },
                                                                    ),
                                                                ),
                                                            },
                                                        ),
                                                        location: 52,
                                                    },
                                                ),
                                            ),
                                        },
                                    ],
                                    from_clause: [
                                        Node {
                                            node: Some(
                                                RangeVar(
                                                    RangeVar {
                                                        catalogname: "",
                                                        schemaname: "",
                                                        relname: "films",
                                                        inh: true,
                                                        relpersistence: "p",
                                                        alias: None,
                                                        location: 59,
                                                    },
                                                ),
                                            ),
                                        },
                                    ],
                                    where_clause: None,
                                    group_clause: [],
                                    group_distinct: false,
                                    having_clause: None,
                                    window_clause: [],
                                    values_lists: [],
                                    sort_clause: [],
                                    limit_offset: None,
                                    limit_count: None,
                                    limit_option: Default,
                                    locking_clause: [],
                                    with_clause: None,
                                    op: SetopNone,
                                    all: false,
                                    larg: None,
                                    rarg: None,
                                },
                            ),
                        ),
                    },
                ),
                replace: true,
                options: [],
                with_check_option: NoCheckOption,
            },
        ),
```

## Additional context

I would expect `Temporary` token to be part of `RangeVar` node but parser still panics so I had to make it part of `ViewStmt`, very similar to what we had with the `As` token being part of `CreateDomainStmt` in #61

### DIFF
--- a/crates/codegen/src/get_node_properties.rs
+++ b/crates/codegen/src/get_node_properties.rs
@@ -439,6 +439,17 @@ fn custom_handlers(node: &Node) -> TokenStream {
                 tokens.push(TokenProperty::from(Token::Or));
                 tokens.push(TokenProperty::from(Token::Replace));
             }
+            if let Some(n) = &n.view {
+                match n.relpersistence.as_str() {
+                    // Permanent
+                    "p" => {},
+                    // Unlogged
+                    "u" => tokens.push(TokenProperty::from(Token::Unlogged)),
+                    // Temporary
+                    "t" => tokens.push(TokenProperty::from(Token::Temporary)),
+                    _ => panic!("Unknown ViewStmt {:#?}", n),
+                }
+            }
         },
         "CreateStmt" => quote! {
             tokens.push(TokenProperty::from(Token::Create));
--- a/crates/parser/src/codegen.rs
+++ b/crates/parser/src/codegen.rs
@@ -124,4 +124,20 @@ mod tests {
             ],
         )
     }
+
+    #[test]
+    fn test_create_view() {
+        test_get_node_properties(
+            "create or replace temporary view comedies as select * from films;",
+            SyntaxKind::ViewStmt,
+            vec![
+                TokenProperty::from(SyntaxKind::Create),
+                TokenProperty::from(SyntaxKind::View),
+                TokenProperty::from(SyntaxKind::As),
+                TokenProperty::from(SyntaxKind::Or),
+                TokenProperty::from(SyntaxKind::Replace),
+                TokenProperty::from(SyntaxKind::Temporary),
+            ],
+        )
+    }
 }


crates/codegen/src/get_node_properties.rs
```diff
@@ -439,6 +439,17 @@ fn custom_handlers(node: &Node) -> TokenStream {
                 tokens.push(TokenProperty::from(Token::Or));
                 tokens.push(TokenProperty::from(Token::Replace));
             }
+            if let Some(n) = &n.view {
+                match n.relpersistence.as_str() {
```
cvng, line 443: Here, I would expect `Temporary` token to be part of `RangeVar` node but parser still panics so I had to make it part of `ViewStmt`, very similar to what we had with the `As` token being part of `CreateDomainStmt` in #61

cc @psteinroe

psteinroe: thats an interesting case! I agree with your reasoning. to give a bit of context. Take the substring `create temporary view comedies` as an example. The `create` and the `view` token should be part of the `ViewStmt` itself, while the `temporary` is definitely part of the `RangeVar` node. We cannot create a valid tree out of this, since we would have to open `RangeVar` at `temporary`, close it to go back to `ViewStmt ` and then open it again to be back at `RangeVar` for `comedies`. So yes, we should pull `temporary` back up into `ViewStmt`. thanks!

psteinroe:
btw @cvng feel free to add your sample statements do `tests/data/statements/valid` to also include them in the e2e tests!

crates/codegen/src/get_node_properties.rs
cvng: it makes sense, thanks for the clarification

cvng:
> btw @cvng feel free to add your sample statements do `tests/data/statements/valid` to also include them in the e2e tests!

I will do, I find it generally more convenient to debug at the e2e layer

PR #66, opened 2023-12-12
minor fixes for easier contrib

cvng:

Fixes various warnings encountered when trying to run the project

- testing requires `cargo +nightly test`
- warnings with `cargo check`
- `parser` lib cannot be used a `dependencies` (proto file path is relative)

## What is the new behavior?

- testing with `cargo test`
- no warnings
- `parser` lib can be used a `dependencies`

## Additional context

This PR can be split into multiple PRs if you see fit @psteinroe

### DIFF
--- a/crates/codegen/src/parser.rs
+++ b/crates/codegen/src/parser.rs
@@ -1,5 +1,6 @@
 use pg_query_proto_parser::ProtoParser;
 use quote::quote;
+use std::{env, path, path::Path};
 
 use crate::{
     get_location::get_location_mod, get_node_properties::get_node_properties_mod,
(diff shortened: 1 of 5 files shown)


crates/codegen/src/parser.rs
```diff
@@ -32,3 +33,12 @@ pub fn parser_mod(_item: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
         #get_nodes
     }
 }
+
+fn proto_file_path() -> path::PathBuf {
```
cvng, line 37: Adapted from https://github.com/cvng/postgres_lsp/blob/minor-fixes/xtask/src/main.rs#L34

psteinroe (approved):
Thanks! ♥️

PR #67, opened 2023-12-13
feat: support `CreateEnumStmt`

cvng:

feat: support `CreateEnumStmt`

Parser panics

## What is the new behavior?

```rust
[2023-12-13T14:27:17Z DEBUG parser::codegen::tests] pg_query_root: Some(
        CreateEnumStmt(
            CreateEnumStmt {
                type_name: [
                    Node {
                        node: Some(
                            String(
                                String {
                                    sval: "status",
                                },
                            ),
                        ),
                    },
                ],
                vals: [
                    Node {
                        node: Some(
                            String(
                                String {
                                    sval: "open",
                                },
                            ),
                        ),
                    },
                    Node {
                        node: Some(
                            String(
                                String {
                                    sval: "closed",
                                },
                            ),
                        ),
                    },
                ],
            },
        ),
    )
```

## Additional context

### DIFF
--- a/crates/parser/src/codegen.rs
+++ b/crates/parser/src/codegen.rs
@@ -66,6 +66,8 @@ mod tests {
 
         debug!("selected node: {:#?}", node_graph[node_index]);
 
+        // note: even though we test for strict equality of the two vectors the order
+        // of the properties does not have to match the order of the tokens in the string
         assert_eq!(node_graph[node_index].properties, expected);
         assert_eq!(node_graph[node_index].properties.len(), expected.len());
     }
--- a/crates/parser/src/codegen.rs
+++ b/crates/parser/src/codegen.rs
@@ -66,10 +66,7 @@ mod tests {
 
         debug!("selected node: {:#?}", node_graph[node_index]);
 
-        assert!(node_graph[node_index]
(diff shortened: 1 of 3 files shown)


crates/parser/src/codegen.rs
```diff
@@ -66,10 +66,7 @@ mod tests {
 
         debug!("selected node: {:#?}", node_graph[node_index]);
 
-        assert!(node_graph[node_index]
-            .properties
-            .iter()
-            .all(|p| { expected.contains(p) }));
+        assert_eq!(node_graph[node_index].properties, expected);
```
cvng, line 69: I don't have a strong opinion here but I find it was easier to debug before #62 for 2 reasons:
- strict assertion - so we cannot have unexpected "extra" nodes
- better assertion message of what is missing in the output

cc @psteinroe

psteinroe: Can you add a comment there that even though we test for strict equality of the two vectors the order of the properties does not have to match the order of the tokens in the string?

cvng: either this, or I just tweak the assertion message - happy to revert if requested

PR #71, opened 2023-12-13
refactor(codegen): use protobuf enums

cvng:

Leverage the fact that most of the `match`ed node properties has a corresponding enum in the `protobuf` definition

Matching is done with raw `i32`

## What is the new behavior?

Matching is done with associated enum

## Additional context

This should improve readability and let us omit some comments since we can now jump to the `protobuf` definition


psteinroe (approved):
You're awesome, thanks!

PR #76, opened 2023-12-14
feat: extend support for create view

cvng:

extend support for `create view`

panics with `with [local | cascaded] check`

## What is the new behavior?

parser returns

## Additional context


crates/codegen/src/get_node_properties.rs
```diff
@@ -385,15 +385,26 @@ fn custom_handlers(node: &Node) -> TokenStream {
...
             if let Some(n) = &n.view {
                 match n.relpersistence.as_str() {
-                    // Permanent
-                    "p" => {},
-                    // Unlogged
-                    "u" => tokens.push(TokenProperty::from(Token::Unlogged)),
                     // Temporary
                     "t" => tokens.push(TokenProperty::from(Token::Temporary)),
-                    _ => panic!("Unknown ViewStmt {:#?}", n.relpersistence),
+                    _ => {},
```
cvng, line 390: only `temporary` apply to views so we just fallback and do nothing

Issue #82, opened 2023-12-17
add snapshots test back to `parser`


psteinroe:
## Planned Work

We already had it in place but dropped it because it's unnecessary now that the parser panics on invalid statements. but turns out it's actually quite useful when reviewing pull requests for #51. we should add it back using [`insta`](https://docs.rs/insta/latest/insta/) as before. we should be able to simply copy it from the git history of the `statement_parser_test` file.

cvng:
Yeah, I was exactly thinking about this, having `insta` back will help making PRs related to #51 less tedious

psteinroe:
> Another great addition could be testing against postgres [regression tests](https://github.com/postgres/postgres/tree/master/src/test/regress/sql) as well

ohh thats a great idea! we can just copy the sql files, or add the entire postgres repo as a submodule to keep it up-to-date. wdyt?

cvng:
I was planning at first to pull the sql files with a tiny script and later add it as a submodule when progress has been made

psteinroe:
@cvng can you open a separate issue for pulling in the regression tests?

cvng:
@psteinroe sure, I already had a few PRs stacked for #51, so I will proceed to pull the postgres tests once those are merged

psteinroe:
> EDIT: also, should we somehow consolidate parser e2e & unit tests? eg. add get_node_properties in the snapshots

the `get_node_properties` tests should rather serve as a help for devs to figure out the right props fast - I would keep them separated.

PR #87, opened 2023-12-18
feat: add snapshots test back

cvng:

follow-up for #82

no snapshots test

## What is the new behavior?

added snapshots test back in [crates/parser/tests/statement_parser_test.rs][1]

## Additional context

With named tests, snapshots will be named like: `create_table@1.snap` (where `1` is the statement number in the file)

@psteinroe lmk what you think about the output

[1]: https://github.com/supabase/postgres_lsp/pull/87/files#diff-b9af3a2376fef543bc5a740ed028bea8b2e7f537923efd553369484ea2db941b


psteinroe (approved):
awesome, LGTM!

PR #88, opened 2023-12-18
feat: add support for create database

cvng:

add support for `create database`

panics

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context


cvng:
not sure if we wait for #87 to have the snapshot in this PR

crates/parser/tests/snapshots/statements/valid/0047@1.snap
```diff
@@ -0,0 +1,86 @@
...
+          Owner@18..23 "OWNER"
+          Whitespace@23..24 " "
+          Ident@24..27 "abc"
+        Whitespace@27..28 " "
+        DefElem@28..38
+          Connection@28..38 "CONNECTION"
+        Whitespace@38..39 " "
+        Limit@39..44 "LIMIT"
+        Whitespace@44..45 " "
+        Iconst@45..46 "5"
```
psteinroe, line 24: These should be part of DefElement

crates/parser/tests/snapshots/statements/valid/0047@4.snap
```diff
@@ -0,0 +1,47 @@
...
+        Create@1..7 "CREATE"
+        Whitespace@7..8 " "
+        Database@8..16 "DATABASE"
+        Whitespace@16..17 " "
+        Ident@17..18 "x"
+        Whitespace@18..19 " "
+        DefElem@19..27
+          Location@19..27 "LOCATION"
+        Whitespace@27..28 " "
+        Default@28..35 "DEFAULT"
```
psteinroe, line 18: Should also be part of DefElement

psteinroe (changes requested):
Can you use the parent of DefElement to fix this?

crates/parser/tests/snapshots/statements/valid/0047@1.snap
cvng: thanks, fixed in 45f88be683ba728c8454d02c6a2feb1349f3ce58

crates/parser/tests/snapshots/statements/valid/0047@4.snap
cvng: fixed in 45f88be683ba728c8454d02c6a2feb1349f3ce58

psteinroe (approved):
lgtm, thanks!

PR #92, opened 2023-12-19
feat: add support for create publication

cvng:
## What kind of change does this PR introduce?

Bug fix, feature, docs update, ...

## What is the current behavior?

Please link any relevant issues here.

## What is the new behavior?

panics

<details>
<summary>View log</summary>

```rust
Parsing node CreatePublicationStmt(
    CreatePublicationStmt {
        pubname: "mypublication",
        options: [],
        pubobjects: [
            Node {
                node: Some(
                    PublicationObjSpec(
                        PublicationObjSpec {
                            pubobjtype: PublicationobjTable,
                            name: "",
                            pubtable: Some(
                                PublicationTable {
                                    relation: Some(
                                        RangeVar {
                                            catalogname: "",
                                            schemaname: "",
                                            relname: "users",
                                            inh: true,
                                            relpersistence: "p",
                                            alias: None,
                                            location: 43,
                                        },
                                    ),
                                    where_clause: None,
                                    columns: [],
                                },
                            ),
                            location: 0,
                        },
                    ),
                ),
            },
            Node {
                node: Some(
                    PublicationObjSpec(
                        PublicationObjSpec {
                            pubobjtype: PublicationobjTable,
                            name: "",
                            pubtable: Some(
                                PublicationTable {
                                    relation: Some(
                                        RangeVar {
                                            catalogname: "",
                                            schemaname: "",
                                            relname: "departments",
                                            inh: true,
                                            relpersistence: "p",
                                            alias: None,
                                            location: 50,
                                        },
                                    ),
                                    where_clause: None,
                                    columns: [],
                                },
                            ),
                            location: 50,
                        },
                    ),
                ),
            },
        ],
        for_all_tables: false,
    },
)
thread 'valid_statements' panicked at crates/parser/src/parse/libpg_query_node.rs:373:13:
assertion `left == right` failed: Tried to start node Node {
    kind: PublicationObjSpec,
    depth: 2,
    properties: [],
    location: Some(
        0,
    ),
} with location 0 but current location is 43
  left: 0
 right: 43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'valid_statements' panicked at crates/parser/tests/statement_parser_test.rs:36:17:
Failed to parse statement 0050:
CREATE PUBLICATION mypublication FOR TABLE users, departments;
```

</details>

## Additional context

failing case:

```sql
CREATE PUBLICATION mypublication FOR TABLE users, departments;
^                                        ^
```

[...]

PR #94, opened 2023-12-21
fix: location and add support for create publication

psteinroe:
includes #92

### DIFF
--- a/crates/codegen/src/get_location.rs
+++ b/crates/codegen/src/get_location.rs
@@ -55,6 +55,22 @@ pub fn get_location_mod(proto_file: &ProtoFile) -> proc_macro2::TokenStream {
                 } else {
                     Some(n.location)
                 },
+                NodeEnum::PublicationObjSpec(n) => {
+                    match &n.pubtable {
+                        Some(pubtable) => match &pubtable.relation {
+                            Some(range_var) => Some(range_var.location),
+                            None => Some(n.location),
+                        },
+                        None => Some(n.location),
+                    }
+                },
+                NodeEnum::BooleanTest(n) => {
+                    if n.arg.is_some() {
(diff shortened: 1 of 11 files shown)


crates/codegen/src/get_location.rs
```diff
@@ -55,6 +55,22 @@ pub fn get_location_mod(proto_file: &ProtoFile) -> proc_macro2::TokenStream {
...
+                        None => Some(n.location),
+                    }
+                },
+                NodeEnum::BooleanTest(n) => {
+                    if n.arg.is_some() {
+                        get_location_internal(&n.arg.as_ref().unwrap().node.as_ref().unwrap())
+                    } else {
+                        Some(n.location)
+                    }
+                },
```
psteinroe, line 73: @cvng fixed location here

crates/codegen/src/get_node_properties.rs
```diff
@@ -692,6 +692,74 @@ fn custom_handlers(node: &Node) -> TokenStream {
...
+                if let Some(NodeEnum::PublicationObjSpec(n)) = &n.node {
+                    match n.pubobjtype() {
+                        protobuf::PublicationObjSpecType::PublicationobjTablesInSchema => {
+                            tokens.push(TokenProperty::from(Token::Tables));
+                            tokens.push(TokenProperty::from(Token::InP));
+                            tokens.push(TokenProperty::from(Token::Schema));
+                        },
+                        _ => {}
+                    }
+                }
```
psteinroe, line 732: I think this can be implemented safer. we should do .find() for all types and add the respective tokens if at least one can be found. feel free to fix this @cvng

cvng: sure, let's have this in a next PR with a `test_create_publication` test

cvng (approved):
LGTM

PR #95, opened 2023-12-23
feat: add support for create table like

cvng:

add support for `create table like`

panics

## What is the new behavior?

parser returns

## Additional context

Last batch of PRs before pulling in the `postgres` regression tests


crates/parser/tests/data/statements/valid/0043.sql
```diff
@@ -0,0 +1,11 @@
+CREATE UNLOGGED TABLE cities (name text, population real, altitude double, identifier smallint, postal_code int, foreign_id bigint);
+/* TODO: CREATE TABLE IF NOT EXISTS distributors (name varchar(40) DEFAULT 'Luso Films', len interval hour to second(3), name varchar(40) DEFAULT 'Luso Films', did int DEFAULT nextval('distributors_serial'), stamp timestamp DEFAULT now() NOT NULL, stamptz timestamp with time zone, "time" time NOT NULL, timetz time with time zone, CONSTRAINT name_len PRIMARY KEY (name, len)); */ SELECT 1;
+/* TODO: CREATE TABLE types (a real, b double precision, c numeric(2, 3), d char(4), e char(5), f varchar(6), g varchar(7)); */ SELECT 1;
```
psteinroe, line 3: @cvng why is this todo?

cvng: I have at least 3 more PRs based on this one and I was trying to keep it smaller

For now each `TODO` is a failing case with either a parser bug (no panics but produces an incorrect tree) or missing node properties, eg `TypeName`

The next PRs should be smaller reproduction cases

PR #97, opened 2023-12-27
feat: support `create conversion`

jasonpanosso:

Support [create conversion](https://www.postgresql.org/docs/current/sql-createconversion.html), chip away at #51

LSP quits with exit code 101 upon attempting to parse `create conversion`

<details>
<summary>LSP Log</summary>

> [START][2023-12-27 10:41:20] LSP logging initiated
> [ERROR][2023-12-27 10:41:20] .../vim/lsp/rpc.lua:675	"rpc"	"/home/jason/.cargo/bin/postgres_lsp" "stderr" "thread 'main' panicked at crates/parser/src/parse/libpg_query_node.rs:148:17: could not find node for token Token { kind: Create, text: \"CREATE\", span: 0..6, token_type: ReservedKeyword } at depth 1
> note: run with RUST_BACKTRACE=1 environment variable to display a backtrace"

</details>

## What is the new behavior?

Handles `create conversion` queries

## Additional context

Is there a file naming convention for the tests in `crates/parser/tests/data/statements`? I see some gaps in the numbers, but I couldn't identify a pattern so I just added a new test at the end.

Also, not sure if checking for `n.func_name.len() > 0` is the best approach for handling adding the `From` token, as I don't understand why `func_name` is a `List*`. Not sure when it would ever not be 0 or 1, would seriously appreciate some insight :)

Awesome project, thx so much for the work!


psteinroe:
hey @jasonpanosso, thanks for the contribution!

> Is there a file naming convention for the tests in crates/parser/tests/data/statements? I see some gaps in the numbers, but I couldn't identify a pattern so I just added a new test at the end.

nope, the gaps just occurred because of pull requests being merged in different orders. you can name them however you like. :) 


> Also, not sure if checking for n.func_name.len() > 0 is the best approach for handling adding the From token, as I don't understand why func_name is a List*

I honestly have no idea. The postgres docs pretty clearly show just a single function. Could you maybe add an extra check and panic of the length of the  vector is > 1?

jasonpanosso:
> nope, the gaps just occurred because of pull requests being merged in different orders. you can name them however you like. :)

Makes sense!

> I honestly have no idea. The postgres docs pretty clearly show just a single function. Could you maybe add an extra check and panic of the length of the vector is > 1?

Strange, must be some weird internal pg stuff.. I pushed a fix that panics when the length of the vec is > 1

PR #99, opened 2023-12-29
feat: support `create transform`

jasonpanosso:

Adds support for [create transform](https://www.postgresql.org/docs/current/sql-createtransform.html)

LSP exits with code 101 upon encountering `create transform`

## What is the new behavior?

Handles `create transform` queries

## Additional context

N/A


psteinroe (approved):
lgtm! Thanks 💚

PR #101, opened 2023-12-30
add test for `SELECT ... WHERE ...`

mobley-trent:
#51


mobley-trent:
@psteinroe I don't know if this is a valid PR 😅. I'm just fishing for features to implement in the parser. But if we're testing all properties then I think this would be valid.

psteinroe:
Every test is valuable!

PR #102, opened 2024-01-02
add support for `SELECT .. ORDER BY ...`

mobley-trent:
#51


psteinroe (changes requested):
This is already supported. See failing test. It's part of the sort by node.

mobley-trent:
It didn't seem to be supported with the `SELECT` statement though. I ran the test initially and the token vector returned by the testing function only had the `SELECT` and `FROM` tokens, the `ORDER` and `BY` tokens were not returned.

psteinroe:
Yes, because the tokens are not supposed to be part of the SelectStmt node. If you instead pass the SortBy node to the test function it will return the tokens. Or just add a select with order by to the e2e test. It will pass.

psteinroe (approved):

PR #104, opened 2024-01-02
test: add support for create table as

cvng:

add tests for `create table as`

parser collects errors after `as` keyword (see snapshots)

## What is the expected behavior?

parser returns without errors

## Additional context

same here, reproduction case cc @psteinroe

### DIFF
--- a/crates/parser/tests/snapshots/statements/valid/0043@8.snap
+++ b/crates/parser/tests/snapshots/statements/valid/0043@8.snap
@@ -1,19 +1,40 @@
 ---
 source: crates/parser/tests/statement_parser_test.rs
-description: "/* TODO: CREATE TABLE films2 AS SELECT * FROM films; */ SELECT 1;"
+description: CREATE TABLE films2 AS SELECT * FROM films;
 ---
 Parse {
-    cst: SourceFile@0..64
-      CComment@0..55 "/* TODO: CREATE TABLE ..."
-      SelectStmt@55..64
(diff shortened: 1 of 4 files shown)


crates/parser/tests/snapshots/statements/valid/0043@8.snap
```diff
@@ -1,19 +1,40 @@
...
-      CComment@0..55 "/* TODO: CREATE TABLE ..."
-      SelectStmt@55..64
-        Select@55..61 "SELECT"
-        Whitespace@61..62 " "
-        ResTarget@62..63
-          AConst@62..63
-            Iconst@62..63 "1"
-        Ascii59@63..64 ";"
+    cst: SourceFile@0..42
+      Create@0..6 "CREATE"
```
psteinroe, line 7: It seems like the statement parser does not pick the root statement up. Can you check that?

cvng: sure, let me try to investigate this one, but I will need some guidance for where to look

psteinroe: fixed it 👍

psteinroe:
@cvng if you add the remaining types (e.g. double precision) we can merge this. thanks!

cvng:
> @cvng if you add the remaining types (e.g. double precision) we can merge this. thanks!

added in 7141a1b

PR #106, opened 2024-01-02
feat: add support for type name

cvng:

add support for "some" `typename`

parser panics

## What is the new behavior?

parser returns

## Additional context

as I understands it, the case that causes panics are: 
- multi-word type name, eg: `timestamp with time zone`
- and interval options, eg: `hour to second(3)`

single-word type name, looks ok but I will have to double-check. cc @psteinroe

### DIFF
--- a/crates/codegen/src/get_node_properties.rs
+++ b/crates/codegen/src/get_node_properties.rs
@@ -869,6 +869,37 @@ fn custom_handlers(node: &Node) -> TokenStream {
                 tokens.push(TokenProperty::from(Token::Function));
             }
         },
+        "TypeName" => quote! {
+            let names = n.names
+                .iter()
+                .filter_map(|n| if let Some(NodeEnum::String(s)) = &n.node { Some(s.sval.clone()) } else { None })
+                .collect::<Vec<_>>();
+
+            if names.len() == 2 && names[0] == "pg_catalog" {
+                match names[1].as_str() {
+                    "interval" => {
+                        tokens.push(TokenProperty::from(Token::To)); //
+                        tokens.push(TokenProperty::from(Token::SecondP)); //
+                    },
+                    "timestamptz" => {
+                        tokens.push(TokenProperty::from(Token::With));
+                        tokens.push(TokenProperty::from(Token::Time));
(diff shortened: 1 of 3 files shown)


psteinroe (commented):
For the type names, you can extend the alias definition. I think it's defined directly in the parser code.

cvng:
> For the type names, you can extend the alias definition. I think it's defined directly in the parser code.

OK, after going back in the git history, it looks like the original parser panics was fixed, so not sure if this if required anymore?

the original error and the failing case:

```sh
could not find node for token Token { kind: To, text: "to", span: 98..100, token_type: ReservedKeyword } at depth 4
```

```sql
CREATE TABLE IF NOT EXISTS distributors (name varchar(40) DEFAULT 'Luso Films', len interval hour to second(3), name varchar(40) DEFAULT 'Luso Films', did int DEFAULT nextval('distributors_serial'), stamp timestamp DEFAULT now() NOT NULL, stamptz timestamp with time zone, "time" time NOT NULL, timetz time with time zone, CONSTRAINT name_len PRIMARY KEY (name, len));
                                                                                                  ^
```

cvng:
base case added in 3aba5fbd904023ed51625480b0ebfb4c76c770f6

psteinroe:
lgtm, thanks!

crates/codegen/src/get_node_properties.rs
```diff
@@ -869,6 +869,51 @@ fn custom_handlers(node: &Node) -> TokenStream {
...
+                        const MINUTE: i32 = 11;
+                        const SECOND: i32 = 12;
+
+                        let fields = &n.typmods.first()
+                            .and_then(|node| node.node.as_ref())
+                            .and_then(|node| if let NodeEnum::AConst(n) = node { n.val.clone() } else { None })
+                            .and_then(|node| if let protobuf::a_const::Val::Ival(n) = node { Some(n.ival) } else { None });
+
+                        if let Some(fields) = fields {
+                            match fields.clone() {
```
cvng, line 892: TODO: add other `fields` from [intervaltypmodout](https://github.com/postgres/postgres/blob/REL_15_STABLE/src/backend/utils/adt/timestamp.c#L1120)

PR #109, opened 2024-01-08
feat: add datatypes aliases

cvng:

copy data types from https://www.postgresql.org/docs/current/datatype.html

missing aliases

## What is the new behavior?

added aliases

## Additional context

as per https://github.com/supabase/postgres_lsp/pull/109#discussion_r1445248208, multi-word type names (eg. `timestamp with time zone`) requires a larger change

### DIFF
--- a/crates/parser/src/parse/libpg_query_node.rs
+++ b/crates/parser/src/parse/libpg_query_node.rs
@@ -412,7 +412,23 @@ impl<'p> LibpgQueryNodeParser<'p> {
 }
 
 /// list of aliases from https://www.postgresql.org/docs/current/datatype.html
-const ALIASES: [&[&str]; 2] = [&["integer", "int", "int4"], &["real", "float4"]];
+const ALIASES: [&[&str]; 15] = [
+    &["bigint", "int8"],
+    &["bigserial", "serial8"],
+    &["bit varying", "varbit"],
+    &["boolean", "bool"],
+    &["character", "char"],
+    &["character varying", "varchar"],
+    &["double precision", "float8"],
+    &["integer", "int", "int4"],
+    &["numeric", "decimal"],
+    &["real", "float4"],
+    &["smallint", "int2"],
+    &["smallserial", "serial2"],
+    &["serial", "serial4"],
+    &["time with time zone", "timetz"],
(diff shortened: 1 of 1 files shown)


crates/parser/src/parse/libpg_query_node.rs
```diff
@@ -412,7 +412,23 @@ impl<'p> LibpgQueryNodeParser<'p> {
...
+    &["character", "char"],
+    &["character varying", "varchar"],
+    &["double precision", "float8"],
+    &["integer", "int", "int4"],
+    &["numeric", "decimal"],
+    &["real", "float4"],
+    &["smallint", "int2"],
+    &["smallserial", "serial2"],
+    &["serial", "serial4"],
+    &["time with time zone", "timetz"],
```
psteinroe, line 429: This will not work, because we are comparing token by token, and this text will be split up over multiplen tokens. It requires a larger change.

psteinroe (changes requested):

crates/parser/src/parse/libpg_query_node.rs
cvng: should this be skipped for now?

psteinroe: I would suggest to skip, bit if you like to get a bit more involved feel free to implement it

crates/parser/src/parse/libpg_query_node.rs
```diff
@@ -412,7 +412,21 @@ impl<'p> LibpgQueryNodeParser<'p> {
...
 
 /// list of aliases from https://www.postgresql.org/docs/current/datatype.html
-const ALIASES: [&[&str]; 2] = [&["integer", "int", "int4"], &["real", "float4"]];
+const ALIASES: [&[&str]; 13] = [
+    &["bigint", "int8"],
+    &["bigserial", "serial8"],
+    &["bit varying", "varbit"],
+    &["boolean", "bool"],
+    &["character", "char"],
+    &["character varying", "varchar"],
```
psteinroe, line 421: These two are also multi word

psteinroe (changes requested):

psteinroe (approved):
lgtm

PR #113, opened 2024-01-13
feat: add support for truncate statement

pert5432:

Progress on https://github.com/supabase/postgres_lsp/issues/51

`TRUNCATE` statements don't get parsed

## What is the new behavior?
`TRUNCATE` statements get parsed

pg_query output: 
```rust
TruncateStmt(
            TruncateStmt {
                relations: [
                    Node {
                        node: Some(
                            RangeVar(
                                RangeVar {
                                    catalogname: "",
                                    schemaname: "",
                                    relname: "users",
                                    inh: true,
                                    relpersistence: "p",
                                    alias: None,
                                    location: 15,
                                },
                            ),
                        ),
                    },
                ],
                restart_seqs: false,
                behavior: DropRestrict,
            },
        )
```

## Additional context
1. I am not sure if the name of the table to be truncated should be a part of the result token stream. For example in `test_create_type` the name of the type is not a part of the token stream but in `test_create_database` the name of the database is.


psteinroe:
Thanks! Can you add a truncate statement to the integration tests too?

pert5432:
I added a snapshot test but it only works with `TRUNCATE table_name` and not `TRUNCATE TABLE table_name` syntax. 
It fails with: 
```
---- valid_statements stdout ----
thread 'valid_statements' panicked at crates/parser/src/parse/libpg_query_node.rs:147:17:
could not find node for token Token { kind: Table, text: "TABLE", span: 9..14, token_type: ReservedKeyword } at depth 2
```
My guess is that its because I don't emit a token for the `TABLE` keyword in the custom handler for `TruncateStmt` but I don't know how to determine whether I should emit that token based on the output of libpg_query which looks like this:
```rust
pub struct TruncateStmt {
    #[prost(message, repeated, tag="1")]
    pub relations: ::prost::alloc::vec::Vec<Node>,
    #[prost(bool, tag="2")]
    pub restart_seqs: bool,
    #[prost(enumeration="DropBehavior", tag="3")]
    pub behavior: i32,
}
```

I would greatly appreciate help resolving this.

psteinroe:
Thanks for looking into this! It seems like [Table is an optional token](https://www.postgresql.org/docs/current/sql-truncate.html). So just pass it. The parser is designed so that it can handle optional tokens.

pert5432:
Thanks for the clarification! Fixed it and added more tests.

PR #118, opened 2024-01-24
fix: support `create type`

cvng:

add support for object type in `DefineStmt`

--

## What is the new behavior?

--

## Additional context

while working on this, I realized there is a few statements that have an AST mismatch from `pg_query`:

https://github.com/search?q=repo%3Asupabase%2Fpostgres_lsp+%22stmts%3A+%5B%5D%2C%22&type=code

with either: no AST (sometimes expected) or AST with wrong nodes (eg. for `create procedure`)

@psteinroe lmk if i should try to investigate these ones next

### DIFF
--- a/crates/parser/src/codegen.rs
+++ b/crates/parser/src/codegen.rs
@@ -273,6 +273,19 @@ mod tests {
     #[test]
     fn test_create_type() {
         test_get_node_properties(
+            "create type type1",
+            SyntaxKind::DefineStmt,
+            vec![
+                TokenProperty::from(SyntaxKind::Create),
+                TokenProperty::from(SyntaxKind::TypeP),
+                TokenProperty::from("type1".to_string()),
+            ],
+        )
+    }
+
+    #[test]
+    fn test_create_composite_type() {
+        test_get_node_properties(
             "create type type1 as (attr1 int4, attr2 bool);",
             SyntaxKind::CompositeTypeStmt,
             vec![
--- a/crates/parser/src/codegen.rs
+++ b/crates/parser/src/codegen.rs
@@ -291,6 +291,7 @@ mod tests {
             vec![
                 TokenProperty::from(SyntaxKind::Create),
                 TokenProperty::from(SyntaxKind::TypeP),
+                TokenProperty::from(SyntaxKind::As),
             ],
         )
     }
--- a/crates/parser/src/parse/statement_start.rs
+++ b/crates/parser/src/parse/statement_start.rs
@@ -191,25 +191,57 @@ pub static STATEMENT_START_TOKEN_MAPS: LazyLock<Vec<HashMap<SyntaxKind, Vec<Toke
             ],
         ));
 
-        // CREATE [ OR REPLACE ] OPERATOR
+        // CREATE OPERATOR
         m.push((
             SyntaxKind::DefineStmt,
             &[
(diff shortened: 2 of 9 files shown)


crates/parser/src/codegen.rs
```diff
@@ -272,6 +272,19 @@ mod tests {
...
+            vec![
+                TokenProperty::from(SyntaxKind::Create),
+                TokenProperty::from(SyntaxKind::TypeP),
+                TokenProperty::from("type1".to_string()),
+            ],
+        )
+    }
+
+    #[test]
+    fn test_create_composite_type() {
```
cvng, line 287: NOTE: `create type as` constitute a composite type

psteinroe:
@cvng thanks! yes, there are some that do not work yet. I have already created an issue to investigate. if you like, please go ahead and check them.

crates/parser/src/parse/statement_start.rs
```diff
@@ -191,25 +191,57 @@ pub static STATEMENT_START_TOKEN_MAPS: LazyLock<Vec<HashMap<SyntaxKind, Vec<Toke
...
-        // CREATE [ OR REPLACE ] TYPE
+        // CREATE TYPE name
         m.push((
             SyntaxKind::DefineStmt,
             &[
                 SyntaxToken::Required(SyntaxKind::Create),
-                SyntaxToken::Optional(SyntaxKind::Or),
-                SyntaxToken::Optional(SyntaxKind::Replace),
                 SyntaxToken::Required(SyntaxKind::TypeP),
+                SyntaxToken::Required(SyntaxKind::Ident),
```
cvng, line 209: the fix part is here: `DefineStmt` was missing `SyntaxKind::Ident`

otherwise I have removed some illegal tokens for some statements (eg. `or replace`) based on the [docs][1]

[1]: https://neon.tech/docs/postgres/sql-createtype

crates/parser/tests/data/statements/valid/0046.sql
```diff
@@ -1,6 +1,6 @@
 CREATE TYPE type1;
 CREATE TYPE type1 AS (attr1 int4, attr2 bool);
-CREATE TYPE type1 AS (attr1 int4 COLLATE collation1, attr2 bool);
+/* TODO: CREATE TYPE type1 AS (attr1 int4 COLLATE collation1, attr2 bool); */ SELECT 1;
```
cvng, line 3: can't get `COLLATE` to work here, I have tried `CollateClause` `ColumnDef` `CompositeTypeStmt` with no luck

it is obfuscated in the macro call but stack trace gives us a hint that it panics in `codegen::get_location_internal`

crates/parser/tests/snapshots/statements/valid/0046@1.snap
```diff
@@ -4,13 +4,39 @@ description: CREATE TYPE type1;
...
+        Create@0..6 "CREATE"
+        Whitespace@6..7 " "
+        TypeP@7..11 "TYPE"
+        Whitespace@11..12 " "
+        Ident@12..17 "type1"
+        Ascii59@17..18 ";"
     ,
     errors: [],
-    stmts: [],
+    stmts: [
```
cvng, line 16: 🎉

psteinroe:
awesome!

PR #141, opened 2024-08-26
adds postgres version to the schema cache.

harshit54:
Adds the postgres version in the schema cache using SQL from the `pg-meta` library.

### DIFF
--- a/crates/pg_schema_cache/src/schema_cache.rs
+++ b/crates/pg_schema_cache/src/schema_cache.rs
@@ -6,6 +6,7 @@ use crate::functions::Function;
 use crate::schemas::Schema;
 use crate::tables::Table;
 use crate::types::PostgresType;
+use crate::versions::Version;
 
 #[derive(Debug, Clone, Default)]
 pub struct SchemaCache {
@@ -13,6 +14,7 @@ pub struct SchemaCache {
     pub tables: Vec<Table>,
     pub functions: Vec<Function>,
     pub types: Vec<PostgresType>,
+    pub versions: Vec<Version>,
 }
 
 impl SchemaCache {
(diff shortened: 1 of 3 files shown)


crates/pg_schema_cache/src/versions.rs
```diff
@@ -0,0 +1,43 @@
...
+  current_setting('server_version_num') :: int8 AS version_num,
+  (
+    select
+      count(*) :: int8 AS active_connections
+    FROM
+      pg_stat_activity
+  ) AS active_connections,
+  current_setting('max_connections') :: int8 AS max_connections;"#
+        )
+        .fetch_all(pool)
```
psteinroe, line 30: why not just one row?

crates/pg_schema_cache/src/schema_cache.rs
```diff
@@ -6,13 +6,15 @@ use crate::functions::Function;
...
 use crate::types::PostgresType;
+use crate::versions::Version;
 
 #[derive(Debug, Clone, Default)]
 pub struct SchemaCache {
     pub schemas: Vec<Schema>,
     pub tables: Vec<Table>,
     pub functions: Vec<Function>,
     pub types: Vec<PostgresType>,
+    pub versions: Vec<Version>,
```
psteinroe, line 17: any reason to have a Vec instead of a single struct?

psteinroe (changes requested):
Thanks for the contribution! Just one minor note 🙏🏼

harshit54:
Hey @psteinroe , thanks for the reply.

While it does make sense to have a single type instead of a Vec, the `SchemaCacheItem` `trait` is defined as 

```rust
pub trait SchemaCacheItem {
    type Item;

    async fn load(pool: &PgPool) -> Vec<Self::Item>;
}
```
is defined as having a `Vec` of the type. Should we make the `version` implement a different trait?

psteinroe:
Ah yeah I think it makes sense to refactor this a bit. Do you want to do it in your pr or should we merge and do a follow-up? 🙏🏼

harshit54:
Let's do a follow up. I'm not sure how the refactoring will look like at the moment.

psteinroe:
sounds good! the ci is failing due to an unrelated issue. I am going to fix it soon and then merge 🙏🏼

PR #142, opened 2024-10-04
naive statement splitter

psteinroe:
naive attempt at splitting statements.

implements the splitter as a simple Pratt Parser. The approach now is the opposite of what I tried before: Instead of assuming it works perfectly as long as we cover all possible edge cases, we assume it does not work at all. We want to make all important cases work (mostly DML, a few DDL statements), while asking the user to fallback to a double newline if they are seeing weird things coming from the lsp. 

- [x] make it work well for all dml statements and all create (or replace) ddl statements


crates/pg_statement_splitter/src/parser.rs
```diff
@@ -33,164 +37,94 @@ pub struct Parse {
...
+        // to make everything simpler, we just filter them out
+        // the token holds the text range, so we dont need to worry about that
+        let tokens = lex(sql)
+            .iter()
+            .filter(|t| {
+                return !WHITESPACE_TOKENS.contains(&t.kind)
+                    || (t.kind == SyntaxKind::Newline && t.text.chars().count() > 1);
+            })
+            .rev()
+            .cloned()
```
juleswritescode, line 51: nit: `into_iter()` instead of `.cloned()` ?

I'm not sure, but if I understand it correctly, we could skip some allocations

crates/pg_statement_splitter/src/parser.rs
```diff
@@ -33,164 +37,94 @@ pub struct Parse {
...
+            .collect::<Vec<_>>();
+
         Self {
-            eof_token: Token::eof(usize::from(tokens.last().unwrap().span.end())),
             ranges: Vec::new(),
+            eof_token: Token::eof(usize::from(tokens.first().unwrap().span.end())),
```
juleswritescode, line 56: will this throw if somebody opens an empty sql file?

crates/pg_statement_splitter/src/parser.rs
```diff
@@ -33,164 +37,94 @@ pub struct Parse {
...
-                })
-                .collect(),
+            ranges: self.ranges,
             errors: self.errors,
         }
     }
 
-    pub fn start_stmt(&mut self) {
+    /// Start statement
+    pub fn start_stmt(&mut self) -> Token {
```
juleswritescode, line 72: Should we return a `Token` here? It might be confusing when a dev works with this and assumes that the next `.peek()` yields a different token

juleswritescode (commented):
Wow, this is so much simpler than the previous implementation. I love it.

crates/pg_statement_splitter/src/parser.rs
psteinroe: this will make `tokens` a reference. but something feels off about this implementation for sure. I would rather use a pointer index instead of popping from the vector. I was struggling to implement this nicely because I wanted the pointer to be an `usize` no matter what. but we need to start the pointer at `-1`, or always subtract 1 when accessing the vector. but I might give it another go.

crates/pg_statement_splitter/src/parser.rs
psteinroe: yes, I am not happy with the start / stop stmt methods yet. they are doing things that you would not assume they would do. right now `start_stmt` `peek`s and sets the peeked tokens start as the current start statement. that feels off for me. do you think that not returning the token would improve this? I was thinking to maybe just have the user pass in the token on which to start as an argument.

psteinroe:
@juleswritescode I refactored the parser to work with an index into the vector. this comes with a few benefits:
- no cloning
- no "save last token" anymore
- just a single pass over tokens since we do not filter and clone on init
- no impact for the user of the parser

crates/pg_statement_splitter/src/parser.rs
psteinroe: I refactored this a bit. see below :)

crates/pg_statement_splitter/src/parser.rs
psteinroe: I refactored this a bit. see below :)

crates/pg_statement_splitter/src/parser/dml.rs
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    statement(p);
+}
+
+pub(crate) fn select(p: &mut Parser) {
+    p.expect(SyntaxKind::Select);
+
+    unknown(p);
+}
```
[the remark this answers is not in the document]
psteinroe: ctes were really straightforward already!

crates/pg_statement_splitter/src/parser.rs
```diff
@@ -46,66 +46,96 @@ impl Parser {
...
-    fn advance(&mut self) -> Token {
-        let token = self.tokens.pop().unwrap_or(self.eof_token.clone());
-
-        self.last_token_end = Some(token.span.end());
-
-        token
+    fn advance(&mut self) -> &Token {
+        let mut first_relevant_token = None;
+        loop {
+            let token = self.tokens.get(self.next_pos).unwrap_or(&self.eof_token);
```
juleswritescode, line 118: self.peek? 🤓

crates/pg_statement_splitter/src/lib.rs
```diff
@@ -1,137 +1,68 @@
...
-                        }
-                    }
-                }
+    use super::*;
 
-                parser.expect(SyntaxKind::Ascii59);
+    struct Tester {
+        input: String,
+        parse: Parse,
+    }
```
juleswritescode, line 26: ah nice, that's convenient

crates/pg_statement_splitter/src/parser.rs
psteinroe: does not work, because advance needs a `&mut self`, while `peek` just takes a `&self`.

(the compiler told me this when I tried it)

psteinroe:
@juleswritescode I think this is ready now. dml statements should work good enough.

crates/pg_statement_splitter/src/parser.rs
```diff
@@ -52,145 +76,116 @@ impl Parser {
...
-            self.error_at_pos(
-                format!("Expected {:#?}, found {:#?}", kind, self.nth(0, false)),
-                self.pos + 1,
-            );
-        }
+
+        self.errors.push(SyntaxError::new(
+            format!("Expected {:#?}", kind),
+            self.peek().span,
+        ));
```
juleswritescode, line 184: very nice! much cleaner.

juleswritescode (approved):
amazing work 💪

PR #144, opened 2024-10-13
fix(vscode): extension command, server crate path

juleswritescode:

When we install and use the VS Code extension, we're now installing the correct crate and running the correct binary.
I also added already-in-use, non-default settings to the `.prettierrc.js` file, so that the diff is more readable.

Running `cargo xtask install` fails because it's looking for a `postgres_lsp` crate; there is none. The crate is called `pg_lsp`.

Starting the VSCode extension fails because it tries to execute the `postgres_lsp` binary; it's called `pglsp`. 

## What is the new behavior?

We're installing the correct crate and executing the correct binary. 

It's working now 🥳🥳🥳:

Screenshot 2024-10-13 at 10 45 31

### DIFF
--- a/editors/code/.prettierrc.js
+++ b/editors/code/.prettierrc.js
@@ -1,7 +1,8 @@
 module.exports = {
-  // use 100 because it's Rustfmt's default
-  // https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#max_width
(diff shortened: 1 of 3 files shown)


psteinroe:
awesome! 🫶🏼

PR #150, opened 2024-12-01
feat(completions): basic scoring algorithm for tables

juleswritescode:
This PR tries to improve the existing `pg_completions` crate – autocompletions! 

The goal was to autocomplete table names based on the input:
1. `select * from us` should put the `users` table before the `emails` table
2. `select * from private.u` should put the `private.uncommon` before the `public.users` table

And whoopdidoopdie, it works.

An overview of the changes:
- Separated the types into `Builder` & `Context` as well as `CompletionItem` + `CompletionRelevance` to rank them (stolen from `rust-analyzer`)
- Removed some unnecessary lifetimes
- The CompletionBuilder now returns the `tower_lsp::CompletionItem` data structure
- Added some integation tests
- Made sure all crates use the workspace's `tower-lsp` version
- Added a small `README.md` to the `pg_completions` crate
- Added a small script that prints the tree-sitter tree of an sql file


crates/pg_completions/Cargo.toml
```diff
@@ -12,6 +12,7 @@ tree-sitter.workspace = true
 tree_sitter_sql.workspace = true
 pg_schema_cache.workspace = true
 pg_test_utils.workspace = true
+tower-lsp.workspace = true
```
psteinroe, line 15: I would prefer to keep language server specifics out of the feature crates. you can find a good rationale on this [here](https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/architecture.md#crateside-crateside-db-crateside-assists-crateside-completion-crateside-diagnostics-crateside-ssr):

> Architecture Invariant: ide crate strives to provide a perfect API. Although at the moment it has only one consumer, the LSP server, LSP does not influence its API design. Instead, we keep in mind a hypothetical ideal client -- an IDE tailored specifically for rust, every nook and cranny of which is packed with Rust-specific goodies.

Biome does the same thing: lsp types are only used within the lsp service. it acts as a client to the workspace (biome_service), which in turn returns its own types. this makes the boundary clearer, and makes sure we are building a postgres tool, and not "just" the language server. especially since we will now have a cli with potentially a `lint` command from the start. the linter should not return lsp-specific `CodeActions` or `Diagnostics`, but just generic lint diagnostics that can also be used by a cli.

psteinroe (changes requested):
very nice! just one thing about the api boundary. thanks for doing this!!

crates/pg_completions/Cargo.toml
juleswritescode: That is a very good point! Adjusted.

I did implement `From<pg_completions::CompletionItem> for lsp_types::CompletionItem` within the crate. It's not possible to implement it in the `pg_lsp` crate – there, both `CompletionItem`s are foreign types.

This means, however, that we still depend on `tower_lsp` in the `pg_completions` crate. What do you think about that?

psteinroe (approved):
Awesome, thanks for working on the feedback!

PR #156, opened 2024-12-14
chore(schema_cache): extract queries to files

juleswritescode:
I started working on function auto-completions and had to read some schema queries.
We could extract them to files, maybe that's easier to work with? Feel free to close the issue if you want to keep the queries in the rust files.


psteinroe:
If the macro still works let's do it! 🙏🏼

PR #157, opened 2024-12-16
fix: small bug in whitespace_tokens, added a few tests for such bugs.

zevaverbach:

Bug fix.

`whitespace_tokens` adds the wrong `kind` attribute for `Token` when the regex has captured a tab (`Newline`)

## What is the new behavior?

It uses the correct `kind` attribute, `Tab`.


psteinroe:
thanks for your contributions!

PR #160, opened 2024-12-16
avoid building skipped tests vector in every loop

zevaverbach:

Small test function optimization.

`SKIPPED_REGRESS_TESTS` are vectorized on every loop (inside two loops).

## What is the new behavior?

The vector is defined before the loops.


psteinroe:
thanks for your contributions!

PR #161, opened 2024-12-17
fix: user's cursor can be offset for autocompletion

juleswritescode:
Suppose we have an incomplete query and the user's cursor is too far off:

`select * from           {}` 

In that case, we couldn't walk the tree-sitter tree.
When we tried to go to the child node at the cursor position, tree-sitter would stay on the `PROGRAM` node (every tree's head) which lead to unexpected contexts and infinite recursion.

That's now fixed! We simply "move the cursor" to the previous node:

`select * from       {}` becomes `select * from{}`.

Note:  Treesitter is already able to deal with ` {}        select * from`.

Some additional stuff:
- added a `tree-print` script to the `justfile` (that's much cooler than Makefile, btw)
- adjusted and used some test-helpers, adjusted the `tree_print` binary to ignore newlines


crates/pg_completions/src/context.rs
```diff
@@ -127,9 +145,14 @@ impl<'a> CompletionContext<'a> {
...
             }
 
+            "keyword_from" => {
+                self.wrapping_clause_type = "keyword_from".try_into().ok();
+            }
+
             _ => {}
         }
 
+        // We have arrived at the leaf node
```
psteinroe, line 155: I love that you also use `we` when writing comments. it's so welcoming to read.

justfile
```diff
@@ -130,6 +130,10 @@ new-crate name:
   cargo new --lib crates/{{snakecase(name)}}
   cargo run -p xtask_codegen -- new-crate --name={{snakecase(name)}}
 
+# Prints the treesitter tree of the given SQL file
+tree-print file:
+	cargo run --bin tree_print -- -f {{file}}
+
```
psteinroe, line 136: love it! will add this alter for the parser too

psteinroe (approved):
great work!!

crates/pg_completions/src/providers/tables.rs
```diff
@@ -76,7 +76,7 @@ mod tests {
             (format!("select * from us{}", CURSOR_POS), "users"),
             (format!("select * from em{}", CURSOR_POS), "emails"),
             // TODO: Fix queries with tree-sitter errors.
```
juleswritescode, line 78:
```suggestion
```

PR #163, opened 2024-12-19
feat(schema_cache): `column` type + query

juleswritescode:
Introduces an all-new, immediately-usable, delivered-straight-to-your-door data type in the `SchemaCache`: the `Column`. 

The next step is to use that data type in the `pg_completions` crate – so we can suggest columns in a select clause, based on the queried table.

I wrote the query myself, with a little bit of help from ChatGPT and the postgres docs, so please make sure it follows best practices. 🙌🏻 

Also wrote and renamed a test, and added full features for tokio on workspace level.


crates/pg_schema_cache/src/queries/columns.sql
```diff
@@ -0,0 +1,60 @@
...
+      -- v: views
+      -- m: materialized views
+      -- f: foreign tables
+      -- p: partitioned tables
+      c.relkind in ('r', 'v', 'm', 'f', 'p')
+  ),
+  available_indexes as (
+    select
+      unnest (ix.indkey) as attnum,
+      ix.indisprimary as is_primary,
```
juleswritescode, line 22: Note that if we're using compounds keys, e.g. `(user_id, tax_number)`, we'd have `is_primary == true` for `tax_number`. 

This might be problematic for index suggestions etc. – wdyt?

psteinroe (commented):
I appreciate the effort to write a custom query, but what didn't you like from this one? :D 

https://github.com/supabase/postgres-meta/blob/master/src/lib/sql/columns.sql

juleswritescode:
Ah, I wasn't aware of the `postgres-meta` repo at all, I only took a look at [this one](https://github.com/supabase/supabase/blob/master/apps/studio/data/database/table-columns-query.ts)...

We can use the `postgres-meta` query if you prefer, it's probably battle-proven. 

But it does lack some info that seemed to me like we might need it in the future: 
- the `type_id` which we could use to find the relevant type on the SchemaCache
- whether the column is a primary key
- the length of a varchar

Of course I'm not sure whether we'll need that data and whether it is fetched 100% correctly. :)

Would you like me to use my query, the `postgres-meta` query, or the `postgres-meta` query with the added properties?

psteinroe:
let's use yours and we keep it in mind when issues come up that it might be worth to go to postgrest-meta for it

PR #164, opened 2025-01-02
feat(completions): add `textDocument/completion` request to LSP + workspace handlers

juleswritescode:
This PR adds the `textDocument/completion` capability to the lsp server 🎉

- added completion to the lsp-server's `ServerCapabilities`
- Removed the `println` statements from the `workspace`, since they were writing to `stdout` and thus led to crashes for the lsp-client
- introduced a couple of `tracing` macros to the lsp server handlers, changed log levels to "info" for methods that happen infrequently (shutdown, initialize etc.)
- upgraded the `vscode-languageclient` to `9.0.1`, which made for better log output

The autocompletions aren't stable right now – completions are sometimes gathered for the wrong statement, and sometimes, the corresponding statement isn't even found. I will fix this in a separate PR 👍

Besides, I wasn't able to start the `pg_cli lsp-proxy` with the VSCode extension. I'll fix this in a separate PR, too.


crates/pg_workspace_new/src/workspace/server.rs
```diff
@@ -342,6 +342,45 @@ impl Workspace for WorkspaceServer {
...
+
+        let schema_cache = self
+            .schema_cache
+            .read()
+            .map_err(|_| WorkspaceError::runtime("Unable to load SchemaCache"))?;
+
+        let result = pg_completions::complete(pg_completions::CompletionParams {
+            position: params.position,
+            schema: &schema_cache,
+            tree: tree.as_deref(),
```
psteinroe, line 378: isn't a ref to the tree sufficient?

crates/pg_workspace_new/src/workspace/server.rs
```diff
@@ -342,6 +342,45 @@ impl Workspace for WorkspaceServer {
...
+        let tree = self.tree_sitter.fetch(&statement_ref);
+        let text = doc
+            .statement_by_id(statement_ref.id)
+            .expect("Found statement_ref but no matching statement")
+            .text;
+
+        let schema_cache = self
+            .schema_cache
+            .read()
+            .map_err(|_| WorkspaceError::runtime("Unable to load SchemaCache"))?;
```
psteinroe, line 373: I agree that the document api is a bit weird, but you should be able to query for `statement_at_offset`, and the `Statement` also includes the `StatementRef`. Open for ideas to improve that api.

crates/pg_workspace_new/src/workspace/server.rs
```diff
@@ -342,6 +342,45 @@ impl Workspace for WorkspaceServer {
...
+
+        let result = pg_completions::complete(pg_completions::CompletionParams {
+            position: params.position,
```
psteinroe, line 376: I think we need to make the position relative to the statement?

crates/pg_lsp_new/src/main.rs
```diff
@@ -0,0 +1,36 @@
+use std::{fs::File, path::PathBuf, str::FromStr};
```
psteinroe, line 1: what benefit does this have over the pg_cli?

psteinroe (changes requested):
a few nits, overall lgtm!

crates/pg_workspace_new/src/workspace/server.rs
juleswritescode: `tree` is an `Option<Arc<Tree>>`:

`&tree` is `&Option<Arc<Tree>>`
`tree.as_ref()` is `Option<&Arc<Tree>>`
`tree.as_deref()` is `Option<&Tree>` 

The way I understand it, `deref` makes it such that, instead of creating a reference to a reference to an i32, we're dereferencing the inner reference, so we get a simple reference to an i32 👍

crates/pg_workspace_new/src/workspace/server.rs
juleswritescode: you're right! thanks. 

Hmm I think the only problem is the name, `StatementRef` sounds to me like a `&Statement` of some kind. 

Maybe the `StatementRef` could be the actual `Statement`, and the doc could have a `get_text(s: &Statement)` method, since the text is probably the only thing that callers care about

crates/pg_workspace_new/src/workspace/server.rs
juleswritescode: I guess you're right, I haven't yet tested this :/

crates/pg_lsp_new/src/main.rs
juleswritescode: I wasn't able to start the extension in vscode with `pg_cli lsp-proxy`, it throws an ENOENT error...

But I'm with you, we shouldn't have a separate main here. I only wanted to use this for debugging, then remove it and fix the extension so it uses the cli in a separate pr 👍

crates/pg_workspace_new/src/workspace/server.rs
psteinroe: Good point! I like that proposal. Will try it out

juleswritescode:
Update: Replacing the VsCode Extension's `OutputChannel` to a `LogOutputChannel` persists the logs, which is quite nice...

If you look at the output, you can see that the returned "JSON" matches our `println` statements. 

I guess `stdout` is sent directly to the client?

Screenshot 2025-01-04 at 10 26 18

Screenshot 2025-01-04 at 10 28 15

psteinroe:
@juleswritescode the cli (and the lsp proxy) writes the output also to a logfile. I am not sure it's the same logs as the lsp client but it helps me debugging a lot. You can find the path to the directory in the pg_fs crate. I also have a script that automatically opens the latest log file (rotated every day, new logfile every hour) and subscribes to it. I can send it over later.

crates/pg_workspace_new/src/workspace/server.rs
```diff
@@ -342,6 +342,57 @@ impl Workspace for WorkspaceServer {
...
+    fn get_completions(
+        &self,
+        params: super::CompletionParams,
+    ) -> Result<pg_completions::CompletionResult, WorkspaceError> {
+        let doc = self
+            .documents
+            .get(&params.path)
+            .ok_or(WorkspaceError::not_found())?;
+
+        let offset = doc.line_and_col_to_offset(params.line, params.column);
```
psteinroe, line 356: this should happen in the lsp layer. we already have helpers for that in `pg_lsp_converters`. the input to the workspace should only be `TextSize` or `TextRange`. the lsp server tracks documents too with the purpose of maintaining the `LineIndex`, which is the data structure we need for the conversion. another reason to handle it on lsp level are different encodings.

crates/pg_workspace_new/src/workspace/server/document.rs
```diff
@@ -81,6 +81,31 @@ impl Document {
             .collect()
     }
 
+    pub fn line_and_col_to_offset(&self, line: u32, col: u32) -> TextSize {
```
psteinroe, line 84: see prev comment - this should be done in the lsp layer, and respective helpers are provided in the `pg_lsp_converters` crate.

editors/code/src/main.ts
```diff
@@ -9,37 +9,39 @@ import {
 
 let client: LanguageClient;
 
-export function activate(_context: ExtensionContext) {
+export async function activate(_context: ExtensionContext) {
     // If the extension is launched in debug mode then the debug server options are used
     // Otherwise the run options are used
     const run: Executable = {
-        command: 'pglsp'
+        command: 'pglsp_new'
```
psteinroe, line 16: can't we just the cli instead with `pg_cli lsp-proxy`?

xtask/src/install.rs
```diff
@@ -137,7 +137,10 @@ fn install_client(sh: &Shell, client_opt: ClientOpt) -> anyhow::Result<()> {
 }
 
 fn install_server(sh: &Shell) -> anyhow::Result<()> {
-    let cmd = cmd!(sh, "cargo install --path crates/pg_lsp --locked --force");
+    let cmd = cmd!(
+        sh,
+        "cargo install --path crates/pg_lsp_new --locked --force"
```
psteinroe, line 142: can't we use `pg_cli` instead? the lsp should not be an entry point.

crates/pg_lsp_new/src/main.rs
psteinroe: ahh saw this comment too late. is this now fixed?

crates/pg_lsp_new/src/server.rs
```diff
@@ -163,19 +163,20 @@ impl LanguageServer for LSPServer {
         self.session.update_all_diagnostics().await;
     }
 
+    #[tracing::instrument(level = "info", skip(self))]
     async fn shutdown(&self) -> LspResult<()> {
         Ok(())
     }
 
-    #[tracing::instrument(level = "trace", skip(self))]
+    #[tracing::instrument(level = "info", skip(self))]
```
psteinroe, line 171: lets undo this once everything is setup

psteinroe (changes requested):
awesome!! left a few comments :)

juleswritescode:
> the cli (and the lsp proxy) writes the output also to a logfile. I am not sure it's the same logs as the lsp client but it helps me debugging a lot. You can find the path to the directory in the pg_fs crate. I also have a script that automatically opens the latest log file (rotated every day, new logfile every hour) and subscribes to it. I can send it over later.

crates/pg_workspace_new/src/workspace/server.rs
juleswritescode: good point, I didn't think about the encoding!

juleswritescode:
@psteinroe I cleaned up the PR, wrote a description and it should now be reviewable 👍
I'll address the other issues we talked about in separate PRs.

psteinroe:
> > the cli (and the lsp proxy) writes the output also to a logfile. I am not sure it's the same logs as the lsp client but it helps me debugging a lot. You can find the path to the directory in the pg_fs crate. I also have a script that automatically opens the latest log file (rotated every day, new logfile every hour) and subscribes to it. I can send it over later.
> 
> Nice! That's useful. I'm still using the `pglsp.log` for server logs. The client logs are different though, those are output by the `vscode-languageclient/node#LanguageClient`, which was the one throwing the Content-Length error.
> 
> Still, please send over the script 🙌

Do not forget to update the username (psteinroe)! 

```sh
# Open latest logfile for the lsp
function open_lsp_log() {
  latest_file=$(ls -t /Users/psteinroe/Library/Caches/dev.supabase-community.pglsp/pglsp-logs | head -n 1)
  if [ -n "$latest_file" ]; then
    tail -f "/Users/psteinroe/Library/Caches/dev.supabase-community.pglsp/pglsp-logs/$latest_file"
  else
    echo 'No log files found'
  fi
}
```

psteinroe (approved):

PR #166, opened 2025-01-06
fix(vscode-ext.): use pg_cli 

juleswritescode:
The VSCode extension now uses the `pg_cli` 🙌


juleswritescode:
The LSP fails immediately when I change a statement, apparently due to an assertion:

Screenshot 2025-01-06 at 11 37 56

psteinroe:
@juleswritescode that's on me, I will look at this this week.

PR #167, opened 2025-01-06
fix: change handling

psteinroe:
iteration on the change handling. after struggling with it for a few days, I simplified it now. its now much simpler and stable, but in certain cases will not be optimal. not optimal means that instead of having one `ModifiedStatement`, we might get up to three `StatementDeleted` + `AddedStatement`. This is relevant for treesitter, where a change can be applied faster if we pass in the prev text and the change. but: since we just parse statement by statement, it should still be fast enough.

Also refactored the Document API to make it much much simpler. Thanks for the input here @juleswritescode!

https://github.com/user-attachments/assets/a8c4d015-80c1-44a4-b0df-43438af35366

todos:
- [x] pr description
- [x] hands on test

Screenshot 2025-01-09 at 10 41 03

9g40bi


crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -362,6 +399,81 @@ mod tests {
...
+                text: ";".to_string(),
+                range: Some(TextRange::new(32.into(), 32.into())),
+            }],
+        };
+
+        let changed2 = d.apply_file_change(&change2);
+        println!("after change 2");
+        println!("{:#?}", d.content);
+        println!("{:#?}", d.statements);
+        println!("{:#?}", changed2);
```
psteinroe, line 440: <insert I am a serious developer clown meme here>

crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -122,13 +122,22 @@ impl Document {
         // special case: if no statement is affected, the affected range is between the prev and
         // the next statement
         if affected.is_empty() {
-            let start = self
+            // since we do not now whether the change should be part of the previous statement, we
+            // will take the range form the start of the previous statement to the start of the
```
psteinroe, line 126: We also need to take the next statement into consideration and then compare the resulting range to check weiter it's the prev or next stmt

crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -54,235 +63,278 @@ impl Document {
...
+        for (index, (_, pos_range)) in self.positions.iter().enumerate() {
+            if pos_range.intersect(change_range).is_some() {
+                affected_indices.push(index);
+                start = start.min(pos_range.start());
+                end = end.max(pos_range.end());
+            } else if pos_range.end() <= change_range.start() {
+                prev_index = Some(index);
+            } else if pos_range.start() >= change_range.end() && next_index.is_none() {
+                next_index = Some(index);
+                break;
```
juleswritescode, line 142: nit: we're implicitly assuming that the `StatementPosition`s are sorted ascending by range. 

If we change the order that the `StatementSplitter` returns statements, this will break.

Should we make the ordering more explicit?

crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -54,235 +63,278 @@ impl Document {
...
-                let adjusted_end = if r.end() >= change.range.unwrap().end() {
-                    if change.is_addition() {
-                        r.end() + change.diff_size()
-                    } else {
-                        r.end() - change.diff_size()
-                    }
+    fn move_ranges(&mut self, offset: TextSize, diff_size: TextSize, is_addition: bool) {
+        self.positions
+            .iter_mut()
+            .skip_while(|(_, r)| offset > r.start())
```
juleswritescode, line 177: We check `offset > r.start()` becuase if the offset is *within* a statements range, that's the modified statement, so we don't have to move it over, right?

crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -54,235 +63,278 @@ impl Document {
...
+
+        // if range is none, we have a full change
+        if change.range.is_none() {
+            return self.apply_full_change(&change.text);
+        }
+
+        // i spent a relatively large amount of time thinking about how to handle range changes
+        // properly. there are quite a few edge cases to consider. I eventually skipped most of
+        // them, because the complexity is not worth the return for now. we might want to revisit
+        // this later though.
```
juleswritescode, line 201: image

juleswritescode (commented):
I still need to takea look at `apply_change` in `change.rs`, but I have to get going now, so here's already a part of the review :)

I love that it's much simpler now!  And I love that auto completion seems to friggin' work.

crates/pg_workspace_new/src/matcher/pattern.rs
```diff
@@ -368,7 +368,7 @@ impl Pattern {
     /// `Pattern` using the default match options (i.e. `MatchOptions::new()`).
     pub fn matches_path(&self, path: &Path) -> bool {
         // FIXME (#9639): This needs to handle non-utf8 paths
-        path.to_str().map_or(false, |s| self.matches(s))
+        path.to_str().is_some_and(|s| self.matches(s))
```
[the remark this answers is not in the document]
psteinroe: Yes! I don't know why they weren't already reported in a prev pr though.

crates/pg_workspace_new/src/workspace/server/change.rs
psteinroe: yes! this is the "easiest" part. anything that's after the changed range will just be moved back. this means that the statement itself stays the same (and the `Statement` which is used to store diagnostics etc). just changing a range is therefore not considered a modification.

crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -54,235 +63,278 @@ impl Document {
...
+        is_addition: bool,
+    ) -> Affected {
+        let mut start = change_range.start();
+        let mut end = change_range.end().min(content_size);
+
+        let mut affected_indices = Vec::new();
+        let mut prev_index = None;
+        let mut next_index = None;
+
+        for (index, (_, pos_range)) in self.positions.iter().enumerate() {
```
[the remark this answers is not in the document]
psteinroe: do you mean bugs within change.rs, or in other parts of the workspace?

crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -461,22 +615,22 @@ mod tests {
 
         assert_eq!(
             changed[0],
-            StatementChange::Deleted(StatementRef {
+            StatementChange::Deleted(Statement {
                 path: path.clone(),
-                id: 0
+                id: 1
```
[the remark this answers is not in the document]
psteinroe: the returned changes are not ordered, just the statements within the document. is this what you mean here?

crates/pg_query_ext/src/lib.rs
```diff
@@ -26,8 +26,7 @@ pub fn parse(sql: &str) -> Result<NodeEnum> {
             .nodes()
             .iter()
             .find(|n| n.1 == 1)
-            .unwrap()
-            .0
-            .to_enum()
-    })
+            .map(|n| n.0.to_enum())
+            .ok_or_else(|| Error::Parse("Unable to find root node".to_string()))
```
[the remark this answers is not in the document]
psteinroe: (I am not even sure why this is an issue. my assumption was that if pg_query parses something it always has at least one node, but that does not seem to be the case.)

crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -3,45 +3,54 @@ use text_size::{TextLen, TextRange, TextSize};
...
 
-impl ChangedStatement {
-    pub fn new_statement(&self) -> Statement {
-        Statement {
-            ref_: self.new_ref.clone(),
-            text: apply_text_change(&self.old.text, Some(self.range), &self.text),
-        }
-    }
+#[derive(Debug, PartialEq, Eq)]
+pub struct ModifiedStatement {
```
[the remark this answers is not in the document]
psteinroe: not explicitly, this was just the result of simplifying the Document api. the enum was there before already. but maybe it did subconsciously. :D

crates/pg_workspace_new/src/workspace/server/change.rs
psteinroe: good point. I added a test to the statement splitter to ensure the output is always sorted. is that sufficient? I think it's easier to make sure it there than to handle it everywhere.

crates/pg_workspace_new/src/workspace/server/document.rs
```diff
@@ -35,119 +28,64 @@ impl Document {
...
-    #[allow(dead_code)]
-    /// Returns the statement ref at the given offset
-    pub fn statement_ref_at_offset(&self, offset: &TextSize) -> Option<StatementRef> {
-        self.statements.iter().find_map(|r| {
-            if r.1.contains(*offset) {
-                Some(self.statement_ref(r))
-            } else {
-                None
-            }
+    pub fn iter_statements(&self) -> impl Iterator<Item = Statement> + '_ {
```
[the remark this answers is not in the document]
psteinroe: yeah, but we create a new `Statement` so we cannot return a ref to it. we might be able to optimise this is caching the Statements but its not worth it I think :D

crates/pg_workspace_new/src/workspace/server/change.rs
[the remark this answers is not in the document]
juleswritescode: Yeah, just wanted to make sure that it's fine that the result is unordered 👍

crates/pg_workspace_new/src/workspace/server/change.rs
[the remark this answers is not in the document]
juleswritescode: within `change.rs` and anything that manages changes. 
It feels to me like mutating an array while iterating over it – but again, just a feeling 👍

crates/pg_workspace_new/src/workspace/server/change.rs
juleswritescode: yeah, great solution 👍

crates/pg_workspace_new/src/workspace/server/change.rs
juleswritescode: gotcha, thanks!

crates/pg_workspace_new/src/workspace/server/change.rs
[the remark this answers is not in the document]
psteinroe: mhh yes I think this entire method can be improved. I didn't find a good way to do "atomic" mutations yet though.

crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -54,235 +69,278 @@ impl Document {
...
+                let new_range = new_ranges[0].add(affected_range.start());
+                let (old_id, old_range) = self.positions[affected_idx];
 
-        self.content = new_content;
+                // move all statements after the afffected range
+                self.move_ranges(old_range.end(), change.diff_size(), change.is_addition());
 
-        self.debug_statements();
+                let new_id = self.id_generator.next();
+                self.positions[affected_idx] = (new_id, new_range);
```
juleswritescode, line 260: beautiful!

crates/pg_workspace_new/src/workspace/server/change.rs
```diff
@@ -54,235 +69,278 @@ impl Document {
...
 
-        changed
-    }
-}
+                changed.push(StatementChange::Modified(ModifiedStatement {
+                    old_stmt: Statement {
+                        id: old_id,
```
juleswritescode, line 264: at this point, that `id` is only stored here, because statement matching that id in the `document` is overwritten, right? 
Do we at some point after this try to find the old statement by id?

juleswritescode (approved):
Great work! Looks much simpler now!

PR #168, opened 2025-01-08
feat(completions): ts_query package, column autocompletion

juleswritescode:
This PR does two things:

1. It adds autocompletion for Columns
2. It adds a `pg_treesitter_queries` crate which cares about extracting useful information from a tree-sitter tree (such as "which tables are mentioned?")

In sum, that makes it such that we have smart autocompletion for columns – see the test cases in `pg_completions/src/complete.rs` for a sneakpeek :)

I chose to put the`pg_treesitter_queries` into their own crate so we can reuse them for `pg_hover` and the likes. 

In addition, I created an `InputQuery` struct for our `test_helpers` which enforces that callers put a `CURSOR_POS` into their queries – I had a nasty bug before because I forgot to do it.


crates/pg_completions/src/complete.rs
```diff
@@ -31,13 +31,14 @@ impl IntoIterator for CompletionResult {
     }
 }
 
-pub fn complete(params: CompletionParams) -> CompletionResult {
+pub fn complete<'a>(params: CompletionParams<'a>) -> CompletionResult {
```
juleswritescode, line 34:
```suggestion
pub fn complete(params: CompletionParams) -> CompletionResult {
```

crates/pg_treesitter_queries/src/queries/relations.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl<'a> QueryTryFrom<'a> for RelationMatch<'a> {
+    type Ref = &'a RelationMatch<'a>;
+}
+
+impl<'a> Query<'a> for RelationMatch<'a> {
+    fn execute(root_node: tree_sitter::Node<'a>, stmt: &'a str) -> Vec<crate::QueryResult<'a>> {
+        let query =
+            tree_sitter::Query::new(tree_sitter_sql::language(), &QUERY).expect("Invalid Query.");
```
juleswritescode, line 61: I tried using a `tokio::sync::OnceCell` here so we would only parse the query once.. but that led to an insurmountable amount of pain and frustration (the `Workspace` trait had to be rewritten)

I would be open to use a `thread_local` here, so at least we'd only have a query per threat. WDYT?

crates/pg_treesitter_queries/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+impl<'a> TreeSitterQueriesExecutor<'a> {
+    pub fn new(root_node: tree_sitter::Node<'a>, stmt: &'a str) -> Self {
+        Self {
+            root_node,
+            stmt,
+            results: vec![],
+        }
+    }
+
+    #[allow(private_bounds)]
```
juleswritescode, line 22: Keeping the `Query` trait private makes it such that we can only ever use the function with the implementations in the `queries` mod, I thought that was a cool pattern

crates/pg_completions/src/context.rs
```diff
@@ -61,18 +70,56 @@ impl<'a> CompletionContext<'a> {
...
+        ctx.gather_info_from_ts_queries();
 
         ctx
     }
 
+    fn gather_info_from_ts_queries(&mut self) {
+        let tree = match self.tree.as_ref() {
+            None => return,
+            Some(t) => t,
+        };
```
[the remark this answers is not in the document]
psteinroe: I I think it's already quite idiomatic like that. only thing that comes to mind is returning an Option<> and then 

 let tree = self.tree.as_ref()?;

but I would keep it like this.

crates/pg_completions/src/providers/columns.rs
```diff
@@ -0,0 +1,114 @@
+use crate::{
+    builder::CompletionBuilder, context::CompletionContext, relevance::CompletionRelevanceData,
+    CompletionItem, CompletionItemKind,
+};
+
+pub fn complete_columns(ctx: &CompletionContext, builder: &mut CompletionBuilder) {
+    let available_columns = &ctx.schema_cache.columns;
```
psteinroe, line 7: is this how others also do it? iterating over all possible options?

crates/pg_completions/src/providers/columns.rs
```diff
@@ -0,0 +1,114 @@
...
+                narrator_id text
+            );
+        "#;
+
+        let queries: Vec<TestCase> = vec![
+            TestCase {
+                message: "correctly prefers the columns of present tables",
+                query: format!(r#"select na{} from public.audio_books;"#, CURSOR_POS),
+                label: "narrator",
+                description: "Table: public.audio_books",
```
psteinroe, line 70: noice!!

crates/pg_treesitter_queries/src/queries/relations.rs
psteinroe: what about a std::sync::LazyLock? but I dont know whether it matter a lot in terms of performance?

psteinroe (approved):
really nice and clean, love it!! answered your questions and left few comments but overall lgtm!

crates/pg_completions/src/context.rs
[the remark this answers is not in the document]
juleswritescode: alright, thank you!

crates/pg_treesitter_queries/src/queries/relations.rs
juleswritescode: Hmm I tried it, and the tests seem to pass. I worry a bit about what happens when we have many concurrent threads, but I guess there'll rarely be more than one autocompletion request, so let's assume it works :)

PR #171, opened 2025-01-14
fix: change at eof

psteinroe:
trying to fix a bug where all statements are deleted when deleting at eof. and when modifying a statement.

fixes:
- using line index of old document during change instead of new
- gracefully handling changes that are out of bounds of the current content. no idea why these are sent, but they are sometimes. instead of investigating where they come from I just added a check to ensure a TextRange is always valid.
- change_range in ModifiedStatement was not adapted to the range of the statement, causing a crash and invalid diagnostics ranges when modifying within a statement.  

ready now. no idea why I didn't finish it earlier. I was sure I already had 😅

### DIFF
--- a/crates/pg_lsp/src/handlers/text_document.rs
+++ b/crates/pg_lsp/src/handlers/text_document.rs
@@ -50,8 +50,6 @@ pub(crate) async fn did_change(
     session: &Session,
     params: lsp_types::DidChangeTextDocumentParams,
 ) -> Result<(), LspError> {
-    tracing::info!("did_change: {:#?}", params);
-
     let url = params.text_document.uri;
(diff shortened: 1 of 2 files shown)


crates/pg_lsp/src/handlers/text_document.rs
```diff
@@ -47,13 +49,15 @@ pub(crate) async fn did_open(
 pub(crate) async fn did_change(
     session: &Session,
     params: lsp_types::DidChangeTextDocumentParams,
-) -> Result<()> {
+) -> Result<(), LspError> {
+    tracing::info!("did_change: {:#?}", params);
```
psteinroe, line 53: will clean this up

PR #172, opened 2025-01-21
cli ergonomics

psteinroe:
goal of this pr is to gear the cli towards the use case we want to support initially: running a check command on migration file(s).

after a discussion with @juleswritescode, this is the plan:

We will have one single `pgtoools check` command. by default, it will scan everything for sql files and lint it. it can be filtered by path, e.g. `pgtools check supabase/migrations` to just lint the migration directory. We will have a `migrations-dir` config option that will enable extra filtering features for files within the directory, primarily `--after` and maybe `--last-nth`. Furthermore, we will also have `--staged` and `--changed` to easily lint just the relevant migrations in a CI environment or a pre-commit hook. The existing `--include` and `--exclude` filters will continue to work. 

To enable `--after`, we will pass a `--migrations-pattern`.

how popular tools store migrations:
- drizzle: migration.sql in a directory with the timestamp and the name. ([link](https://orm.drizzle.team/docs/migrations))
- prisma: same as drizzle ([link](https://www.prisma.io/docs/orm/prisma-migrate/workflows/customizing-migrations))
- supabase: migrations dir with sql files
- typeorm: stores migrations in ts files -> not supported for now

UPDATE: I removed `--migrations-pattern` again - we just try both and check if any yield an expected value. Added `migration_dir` as well as `after` :) 

I initially tried to put it into `get_files_to_process` but that only looks at the inputs from the user and we need the traversal output to look at each file recursively. the right place for that is `can_handle`, which calls `is_path_ignored` from the workspace. the downside of my approach is that we check ignore patterns also for files, whereas before we just checked it for directories and symlinks.... an alternative would be a new workspace api aka `is_migration_ignored`, but that would conflict with `is_path_ignored` imo.

### DIFF
--- a/crates/pg_cli/src/commands/check.rs
+++ b/crates/pg_cli/src/commands/check.rs
@@ -1,5 +1,6 @@
 use crate::cli_options::CliOptions;
 use crate::{CliDiagnostic, Execution, TraversalMode};
+use biome_deserialize::Merge;
 use pg_configuration::PartialConfiguration;
 use pg_console::Console;
 use pg_fs::FileSystem;
@@ -9,9 +10,6 @@ use std::ffi::OsString;
 use super::{get_files_to_process_with_cli_options, CommandRunner};
 
 pub(crate) struct CheckCommandPayload {
-    pub(crate) write: bool,
-    pub(crate) fix: bool,
-    pub(crate) unsafe_: bool,
     pub(crate) configuration: Option<PartialConfiguration>,
     pub(crate) paths: Vec<OsString>,
     pub(crate) stdin_file_path: Option<String>,
@@ -26,12 +24,20 @@ impl CommandRunner for CheckCommandPayload {
     fn merge_configuration(
         &mut self,
         loaded_configuration: LoadedConfiguration,
-        fs: &DynRef<'_, dyn FileSystem>,
-        console: &mut dyn Console,
+        _fs: &DynRef<'_, dyn FileSystem>,
+        _console: &mut dyn Console,
     ) -> Result<PartialConfiguration, WorkspaceError> {
-        let LoadedConfiguration { configuration, .. } = loaded_configuration;
+        let LoadedConfiguration {
+            configuration: mut fs_configuration,
+            ..
+        } = loaded_configuration;
 
-        Ok(configuration)
+        if let Some(configuration) = self.configuration.clone() {
+            // overwrite fs config with cli args
+            fs_configuration.merge_with(configuration);
+        }
+
+        Ok(fs_configuration)
     }
 
     fn get_files_to_process(
@@ -39,7 +45,6 @@ impl CommandRunner for CheckCommandPayload {
         fs: &DynRef<'_, dyn FileSystem>,
         configuration: &PartialConfiguration,
     ) -> Result<Vec<OsString>, CliDiagnostic> {
-        // update this to find migration files
         let paths = get_files_to_process_with_cli_options(
             self.since.as_deref(),
             self.changed,
@@ -57,7 +62,7 @@ impl CommandRunner for CheckCommandPayload {
     }
 
     fn should_write(&self) -> bool {
-        self.write || self.fix
+        false
     }
 
     fn get_execution(
--- a/crates/pg_cli/src/commands/check.rs
+++ b/crates/pg_cli/src/commands/check.rs
@@ -39,6 +39,7 @@ impl CommandRunner for CheckCommandPayload {
         fs: &DynRef<'_, dyn FileSystem>,
         configuration: &PartialConfiguration,
     ) -> Result<Vec<OsString>, CliDiagnostic> {
+        // update this to find migration files
         let paths = get_files_to_process_with_cli_options(
             self.since.as_deref(),
             self.changed,
--- a/crates/pg_cli/src/execute/traverse.rs
+++ b/crates/pg_cli/src/execute/traverse.rs
@@ -481,7 +481,10 @@ impl TraversalContext for TraversalOptions<'_, '_> {
         }
 
         // only allow .sql and .pg files for now
-        if path.extension().is_none_or(|ext| ext != "sql" && ext != "pg") {
+        if path
+            .extension()
+            .is_none_or(|ext| ext != "sql" && ext != "pg")
+        {
             return false;
         }
 
--- a/crates/pg_cli/src/execute/traverse.rs
+++ b/crates/pg_cli/src/execute/traverse.rs
@@ -456,7 +456,13 @@ impl TraversalContext for TraversalOptions<'_, '_> {
 
     fn can_handle(&self, pglsp_path: &PgLspPath) -> bool {
         let path = pglsp_path.as_path();
-        if self.fs.path_is_dir(path) || self.fs.path_is_symlink(path) {
+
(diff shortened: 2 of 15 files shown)


crates/pg_cli/src/execute/traverse.rs
```diff
@@ -456,7 +456,13 @@ impl TraversalContext for TraversalOptions<'_, '_> {
...
     fn can_handle(&self, pglsp_path: &PgLspPath) -> bool {
         let path = pglsp_path.as_path();
-        if self.fs.path_is_dir(path) || self.fs.path_is_symlink(path) {
+
+        let is_valid_file = self.fs.path_is_file(path)
+            && path
+                .extension()
+                .is_some_and(|ext| ext == "sql" || ext == "pg");
+
+        if self.fs.path_is_dir(path) || self.fs.path_is_symlink(path) || is_valid_file {
```
psteinroe, line 465: here is the relevant part in `can_handle`

crates/pg_workspace/src/workspace/server.rs
```diff
@@ -150,13 +157,31 @@ impl WorkspaceServer {
...
+    }
+
     /// Check whether a file is ignored in the top-level config `files.ignore`/`files.include`
     fn is_ignored(&self, path: &Path) -> bool {
         let file_name = path.file_name().and_then(|s| s.to_str());
         // Never ignore PGLSP's config file regardless `include`/`ignore`
         (file_name != Some(ConfigName::pglsp_toml())) &&
             // Apply top-level `include`/`ignore
-            (self.is_ignored_by_top_level_config(path))
+            (self.is_ignored_by_top_level_config(path) || self.is_ignored_by_migration_config(path))
```
psteinroe, line 181: `is_ignored` is called by `is_path_ignored` and we add the new check here.

crates/pg_workspace/src/workspace/server/migration.rs
```diff
@@ -0,0 +1,116 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs;
+    use std::path::PathBuf;
+    use tempfile::TempDir;
+
+    fn setup() -> TempDir {
+        TempDir::new().expect("Failed to create temp dir")
```
psteinroe, line 62: tempfile is required to have proper test that work with `canonicalize`

pglsp.toml
```diff
@@ -13,6 +13,10 @@ username = "postgres"
 password = "postgres"
 database = "postgres"
 
+[migrations]
+migrations_dir = "migrations"
+after = 20230912094327
```
psteinroe, line 18: tested it locally but removed the migrations dir again

crates/pg_cli/src/commands/check.rs
```diff
@@ -56,7 +62,7 @@ impl CommandRunner for CheckCommandPayload {
     }
 
     fn should_write(&self) -> bool {
-        self.write || self.fix
+        false
```
juleswritescode, line 65: sanity check: did you just change this for debugging?

crates/pg_configuration/src/migrations.rs
```diff
@@ -0,0 +1,17 @@
...
+use serde::{Deserialize, Serialize};
+
+/// The configuration of the filesystem
+#[derive(Clone, Debug, Deserialize, Eq, Partial, PartialEq, Serialize, Default)]
+#[partial(derive(Bpaf, Clone, Eq, PartialEq, Merge))]
+#[partial(serde(rename_all = "snake_case", default, deny_unknown_fields))]
+pub struct MigrationsConfiguration {
+    /// The directory where the migration files are stored
+    #[partial(bpaf(long("migrations-dir")))]
+    pub migrations_dir: String,
```
juleswritescode, line 12: I don't know if we use [bpaf-derive](https://crates.io/crates/bpaf_derive/0.1.0), but if we do, we *could* also parse into a `PathBuf`. 

Examples of that were hard to find, so maybe that's not the recommended way – I'm not sure

crates/pg_workspace/src/workspace/server.rs
juleswritescode: very sweet!

crates/pg_workspace/src/workspace/server.rs
```diff
@@ -150,13 +157,28 @@ impl WorkspaceServer {
...
+            .migrations
+            .as_ref()
+            .and_then(|migration_settings| {
+                let ignore_before = migration_settings.after.as_ref()?;
+                let migrations_dir = migration_settings.path.as_ref()?;
+                let migration = migration::get_migration(path, migrations_dir)?;
+
+                Some(&migration.timestamp <= ignore_before)
+            })
+            .unwrap_or(false)
```
juleswritescode, line 172: What happens if we pass a migrations_dir but no `after`? Wouldn't the `and_then` return `None`, so the function would always return `false`?

juleswritescode (approved):
Very beautiful PR and great description! Love it 😍

crates/pg_cli/src/commands/check.rs
psteinroe: no, I removed all the `--fix` stuff because ware not fixing anything. its a leftover from biomes `check` command.

pglsp.toml
psteinroe: good point :D

crates/pg_workspace/src/workspace/server/migration.rs
```diff
@@ -0,0 +1,116 @@
...
+        assert!(migration.is_some());
+        let migration = migration.unwrap();
+        assert_eq!(migration.timestamp, 1234567890);
+        assert_eq!(migration.name, "create_users");
+    }
+
+    #[test]
+    fn test_get_migration_non_migration_file() {
+        let migrations_dir = PathBuf::from("/tmp/migrations");
+        let path = migrations_dir.join("not_a_migration.sql");
```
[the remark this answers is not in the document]
psteinroe: it doesnt match the pattern since it doesnt contain a timestamp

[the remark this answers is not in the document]
psteinroe: clarified it in the test name 👍

crates/pg_workspace/src/workspace/server/migration.rs
```diff
@@ -0,0 +1,116 @@
...
+        return None;
+    }
+
+    // Try Root pattern
+    let root_migration = path
+        .file_name()
+        .and_then(|os_str| os_str.to_str())
+        .and_then(|file_name| {
+            let mut parts = file_name.splitn(2, '_');
+            let timestamp = parts.next()?.parse().ok()?;
```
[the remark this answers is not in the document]
psteinroe: yes, added comments

crates/pg_configuration/src/migrations.rs
psteinroe: that would be preferred, but `biome`s `Merge` does not work with `PathBuf` :(

crates/pg_workspace/src/workspace/server.rs
psteinroe: yes, and we do not want to ignore any migration file if no `after` is specified do we?

PR #186, opened 2025-01-30
feat: release pipeline

juleswritescode:
A hand-made release pipeline that will do the following on manual trigger:

- Run Tests on Windows, MacOS, Linux
- build the `pg_cli` for six rust target-triples
- increment the release-tag based on our commit history
- create a GH release with that tag 
- attach a autogenerated changelog to the release
- attach the compiled binaries to the release

The result for the first release will look something like below – which is why we should use semantic commits and squash merges 🤣:

-- image removed


psteinroe:
@juleswritescode can we write the first release manually? 😂

juleswritescode:
> @juleswritescode can we write the first release manually? 😂

Let's do that :) I'll run it again on my private repo to see if everything works first.
Then, is it fine for you if we just edit the release body? Or should I disable setting the output in the yaml file before we merge?

psteinroe:
I don't know how it works exactly. I fine with anything as long as we can edit the body manually before releasing.

juleswritescode:
Update: the Workflow works with all six runner-types. :)

We can simply publish a release and then edit the Release body and write whatever we want. 

It's ready for review 👍

Screenshot 2025-01-31 at 14 00 16

psteinroe (approved):
Awesome!!

juleswritescode:
This will now create a Draft PR that is only visible to maintainers of the repo. 
The workflow will output a Link to the Draft Release as a Summary. 👍

PR #187, opened 2025-01-30
fix: handle unicode in lexer

### LINKED ISSUE #184: lexer: No token found at position
the error was thrown in `pg_lexer` when testing #172 on [this file](https://github.com/psteinroe/supabase-cache-helpers/blob/6d84a63dd63cc0d636dbd4f3d188d9ac8ad460aa/supabase/migrations/1660333326_test_schema.sql). we should fix this.

psteinroe:
fixes #184 

also using `TextSize` now. but we might want to revisit the lexer at some point because I dont think we should store the text in the tokens anymore.

### DIFF
--- a/crates/pg_lexer/src/lib.rs
+++ b/crates/pg_lexer/src/lib.rs
@@ -3,7 +3,7 @@ mod codegen;
 use pg_query::protobuf::{KeywordKind, ScanToken};
 use regex::Regex;
 use std::{collections::VecDeque, sync::LazyLock};
-use text_size::{TextRange, TextSize};
+use text_size::{TextLen, TextRange, TextSize};
 
 pub use crate::codegen::SyntaxKind;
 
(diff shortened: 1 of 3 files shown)


crates/pg_lexer/src/lib.rs
```diff
@@ -119,25 +119,24 @@ pub fn lex(text: &str) -> Vec<Token> {
...
-                .take(
-                    usize::try_from(pg_query_token.end).unwrap()
-                        - usize::try_from(pg_query_token.start).unwrap(),
-                )
-                .collect();
-            let len = token_text.len();
+
+            // the lexer returns byte indices, so we need to slice
+            let token_text = &text[usize::try_from(pg_query_token.start).unwrap()
+                ..usize::try_from(pg_query_token.end).unwrap()];
```
psteinroe, line 132: the important change is this

juleswritescode (approved):
Nice!

PR #188, opened 2025-01-30
chore: upgrade pg query and drop deprecated crates 

psteinroe:
upgrade `pg_query.rs` to latest.
also:
- set branch for `libpg_query` submodule to `17-latest` and updated the commit
- updated the commit for the `tree-sitter-sql` submodule to latest
- re-activate windows tests 
- adds support for `\c\r` to fix lexer tests on windows (til: https://dev.to/pieter/why-windows-uses-rn-newlines-instead-of-n-126l)

ahhh windows is still failing. I will just drop `pg_syntax` - its not worth to fix the parser now :D


juleswritescode (approved):
Very nice!!  :) 
Should we pin the `pg_query` version on workspace level?

psteinroe:
> Very nice!! :) Should we pin the `pg_query` version on workspace level?

good point - done! :)

PR #189, opened 2025-01-31
docs(pg_lexer): add some info about libpg_query and SyntaxKind

juleswritescode:
I tried to understand how the `pg_lexer` crate and its interaction with `libpg_query` works and thought it would be a good idea to write it down. 

Please check if I understand it correctly.


psteinroe (commented):
Thanks!

psteinroe (approved):

PR #194, opened 2025-02-03
fix(workspace): gracefully handle failing db connections

juleswritescode:
With this PR, we should now be able to run basic diagnostics, even if the workspace can't connect to the Db. I tested it locally with `pg_cli check test.sql` and it seems to work.

The basic gist:

Whenever we called `WorkspaceServer::update_settings`, we would configure the `DatabaseConnection` for *lazy-loading*, but *eagerly* loaded the `SchemaCache`. 

I thought it was weird that we queried the Db in a `update_settings` call, so I refactored the SchemaCache to be lazy-loaded as well. 

For this, I wrapped the whole thing into a `SchemaCacheManager` which checks if we've already cached the requested `PgPool`'s schema, returning guarded handles to the underlying `SchemaCache`. 

### Additional Things
- moved `SchemaCacheManager`, `DbConnection`, and the async runtime helpers into their own modules
- DbConnection::get_pool() will now panic if the developer hasn't set the DbConnection options
- the `completions` handler will now return an empty array if we can't connect to the db
- added an optional `conn_timeout_secs` config option, which is by default set to 10 seconds
- removed SchemaCache::new(), since that only delegated to its `Default` impl

### DIFF
--- a/crates/pg_cli/src/cli_options.rs
+++ b/crates/pg_cli/src/cli_options.rs
@@ -18,7 +18,7 @@ pub struct CliOptions {
     #[bpaf(long("use-server"), switch, fallback(false))]
     pub use_server: bool,
 
-    /// Skip over files containing syntax errors instead of emitting an error diagnostic.
+    /// Skip connecting to the database and only run checks that don't require a database connection.
     #[bpaf(long("skip-db"), switch, fallback(false))]
     pub skip_db: bool,
 
--- a/crates/pg_cli/src/cli_options.rs
+++ b/crates/pg_cli/src/cli_options.rs
@@ -18,6 +18,10 @@ pub struct CliOptions {
     #[bpaf(long("use-server"), switch, fallback(false))]
     pub use_server: bool,
 
+    /// Skip over files containing syntax errors instead of emitting an error diagnostic.
+    #[bpaf(long("skip-db"), switch, fallback(false))]
+    pub skip_db: bool,
+
     /// Print additional diagnostics, and some diagnostics show more information. Also, print out what files were processed and which ones were modified.
     #[bpaf(long("verbose"), switch, fallback(false))]
     pub verbose: bool,
--- a/crates/pg_configuration/src/database.rs
+++ b/crates/pg_configuration/src/database.rs
@@ -28,8 +28,8 @@ pub struct DatabaseConfiguration {
     pub database: String,
 
     /// The connection timeout in seconds.
-    #[partial(bpaf(long("conn_timeout")))]
-    pub conn_timeout_secs: Option<u16>,
+    #[partial(bpaf(long("conn_timeout_secs"), fallback(Some(10)), debug_fallback))]
+    pub conn_timeout_secs: u16,
 }
 
 impl Default for DatabaseConfiguration {
@@ -40,7 +40,7 @@ impl Default for DatabaseConfiguration {
             username: "postgres".to_string(),
             password: "postgres".to_string(),
             database: "postgres".to_string(),
-            conn_timeout_secs: Some(10),
+            conn_timeout_secs: 10,
         }
     }
 }
--- a/crates/pg_configuration/src/database.rs
+++ b/crates/pg_configuration/src/database.rs
@@ -26,6 +26,10 @@ pub struct DatabaseConfiguration {
     /// The name of the database.
     #[partial(bpaf(long("database")))]
     pub database: String,
+
+    /// The connection timeout in seconds.
+    #[partial(bpaf(long("conn_timeout")))]
+    pub conn_timeout: Option<u16>,
 }
 
 impl Default for DatabaseConfiguration {
@@ -36,15 +40,7 @@ impl Default for DatabaseConfiguration {
             username: "postgres".to_string(),
             password: "postgres".to_string(),
             database: "postgres".to_string(),
+            conn_timeout: Some(10),
         }
-    }
-}
-
-impl DatabaseConfiguration {
-    pub fn to_connection_string(&self) -> String {
-        format!(
-            "postgres://{}:{}@{}:{}/{}",
-            self.username, self.password, self.host, self.port, self.database
-        )
     }
 }
--- a/crates/pg_configuration/src/database.rs
+++ b/crates/pg_configuration/src/database.rs
@@ -29,7 +29,7 @@ pub struct DatabaseConfiguration {
 
     /// The connection timeout in seconds.
     #[partial(bpaf(long("conn_timeout")))]
-    pub conn_timeout: Option<u16>,
+    pub conn_timeout_secs: Option<u16>,
 }
 
 impl Default for DatabaseConfiguration {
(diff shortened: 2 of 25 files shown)


crates/pg_workspace/src/workspace/server.rs
```diff
@@ -358,51 +300,51 @@ impl Workspace for WorkspaceServer {
...
-                                    .map(|d| {
-                                        let r = d.location().span.map(|span| span + range.start());
-
-                                        d.with_file_path(path.as_path().display().to_string())
-                                            .with_file_span(r.unwrap_or(range))
-                                    })
-                                } else {
-                                    None
-                                }
+            let pool = connection.get_pool();
```
juleswritescode, line 303: all of the below is just moved by a tab, because I removed the `if let Some(..) = ..` closure

crates/pg_configuration/src/database.rs
```diff
@@ -36,15 +40,7 @@ impl Default for DatabaseConfiguration {
             username: "postgres".to_string(),
             password: "postgres".to_string(),
             database: "postgres".to_string(),
+            conn_timeout_secs: Some(10),
```
psteinroe, line 43: 10 seconds is a long time :D usually the connection will happen either within <1s or never, maybe we should set this a bit lower to report issues earlier?

crates/pg_schema_cache/src/schema_cache.rs
```diff
@@ -42,6 +41,30 @@ impl SchemaCache {
         })
     }
 
+    fn pool_to_conn_str(pool: &PgPool) -> String {
```
psteinroe, line 44: mhh Connection Str Logic and cache keys on the schema cache feels like a leaking abstraction... why not handling this on the schema cache manager itself? After all, it should manage the schema cache for the workspace and can also be knowledgable of the connection string and handle caching?

psteinroe (commented):
Really nice! Thanks for the cleanup too. Left two comments, let's discuss them in the respective threads 🙏🏼

crates/pg_configuration/src/database.rs
psteinroe: I was thinking 2s maybe

juleswritescode: Hmm are you sure? The default SQLX timeout is 30s, and the default mongo driver timeout is 10s. I would expect that they choose sane defaults. Also, what if connection is not done locally, but e.g. from a GH Action Runner to an org's DB, or from a local VS Code to a remote DB?  

I'm happy to go with the 1-2s, just challenging. :)

crates/pg_schema_cache/src/schema_cache.rs
juleswritescode: I had that first, but then I had to maintain another `conn_str: RwLock<String>`, which also felt odd 🤔 I can replicate that in a commit so you can check what that would look like.

My thoughts were that the schema-cache *should* theoretically be the one to know from which DB it pulled the schema. 

Maybe we could form a separate cache-str format for the `SchemaCache`, this doesn't have to be the connection string?

crates/pg_configuration/src/database.rs
psteinroe: Mh yeah good point. I am just worried that users might wonder what's wrong when they won't see any info for 10s. But we can also show something to the user somehow.

crates/pg_schema_cache/src/schema_cache.rs
psteinroe: yeah, maybe use an inner struct on the manager? This way both the connection string and the data is locked together.

juleswritescode:
@psteinroe 

Regarding the "slowness": 

I added `--skip-db` connection that will make it such we never connect to any db. The `DbConnection.get_pool()` will simply return `None`. 

Regarding the timeout:

I tried to add a deferred log to the console, something that would print "Your connection takes longer than expected…" if it takes >=2s – but that required wrapping the `app.console` in a mutex. 

So instead, I added a `user_hints` vector that collects hints to print just after the diagnostics summary. 
(We could reuse this e.g. when a user does not properly separate their statement with newlines for the statement splitter.)

I also added an AtomicBool that indicates whether we were able to get a `PgPool`. If not, we'll print another hint: "Skipped all checks requiring a database connection".

I'm not 100% happy with the result; especially the bool feels inelegant. So hit me with your feedback!

crates/pg_schema_cache/src/schema_cache.rs
juleswritescode: done and much better!

crates/pg_cli/src/cli_options.rs
```diff
@@ -18,6 +18,10 @@ pub struct CliOptions {
     #[bpaf(long("use-server"), switch, fallback(false))]
     pub use_server: bool,
 
+    /// Skip over files containing syntax errors instead of emitting an error diagnostic.
```
psteinroe, line 21: Maybe I am misunderstanding, but is this the correct comment? I thought it would simply not connect to the db and still provide everything we can like lint and Syntax diagnostics

psteinroe (commented):
I like the implementation a lot, but the entire user hint story feels like bloat? For now, I think we should either skip this "feedback" entirely or just add it as another "warning" diagnostics? but not sure

crates/pg_cli/src/cli_options.rs
juleswritescode: no, sorry, that's just a copy-paste error! good catch.

juleswritescode:
> I like the implementation a lot, but the entire user hint story feels like bloat? For now, I think we should either skip this "feedback" entirely or just add it as another "warning" diagnostics? but not sure

juleswritescode:
@psteinroe Also, you're still on vacation for a day, so don't you dare review before that

crates/pg_configuration/src/database.rs
```diff
@@ -26,6 +26,10 @@ pub struct DatabaseConfiguration {
     /// The name of the database.
     #[partial(bpaf(long("database")))]
     pub database: String,
+
+    /// The connection timeout in seconds.
+    #[partial(bpaf(long("conn_timeout")))]
+    pub conn_timeout_secs: Option<u16>,
```
psteinroe, line 32: Not sure, but maybe we can set the default via bpaf?

psteinroe: If not, we should add it to the comments.

psteinroe (approved):
(I am on German grounds again!)

Lgtm! Just have one comment about the timeout 🙏🏼

crates/pg_configuration/src/database.rs
juleswritescode: I played around with using a bpaf fallback, but the `partial` macro requires that the fallback is wrapped in an Option.

If we want to use `display_fallback`, we would also need to implement Display for `Option<u16>`, but we can't because of the Orphan rule.

A quick fix is to add the fallback to the comment, wdyt?

juleswritescode: Ah, we can use the `debug` representation. Works for me, too.

Screenshot 2025-02-09 at 18 27 12

PR #195, opened 2025-02-09
feat(analyser): banDropNotNull rule

juleswritescode:
I tried adding a new lint rule, but the `just gen-lint` script fails:

Screenshot 2025-02-09 at 09 46 28

I can't see a syntax error in this statement:
`alter table users alter column email drop not null;` 

so maybe something is broken?

### DIFF
--- a/crates/pg_analyser/CONTRIBUTING.md
+++ b/crates/pg_analyser/CONTRIBUTING.md
@@ -17,12 +17,14 @@ We follow a naming convention according to what the rule does:
 1. Forbid a concept
 
    ```block
-   no<Concept>
+   ban<Concept>
    ```
 
-   When a rule's sole intention is to **forbid a single concept** the rule should be named using the `no` prefix.
+   When a rule's sole intention is to **forbid a single concept** the rule should be named using the `ban` prefix.
 
-1. Mandate a concept
+   Example: "banDropColumn"
+
+2. Mandate a concept
 
    ```block
    use<Concept>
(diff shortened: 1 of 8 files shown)


crates/pg_analyser/CONTRIBUTING.md
```diff
@@ -246,7 +251,7 @@ The documentation needs to adhere to the following rules:
 
 Here's an example of how the documentation could look like:
 
-```rust
+````rust
```
juleswritescode, line 254: will check this, too – my local prettier adds an extra ` for rust

psteinroe:
Thanks for working through the guide! I will check what's wrong. We should also add a test framework (like biome) to quickly run the rules over a set of sql files and compare expected results. 🙏🏼

psteinroe:
This could even be an issue in the statement splitter 😅 it seems like we are cutting the statement too early.

juleswritescode:
> This could even be an issue in the statement splitter 😅 it seems like we are cutting the statement too early.

juleswritescode:
> Thanks for working through the guide! I will check what's wrong. We should also add a test framework (like biome) to quickly run the rules over a set of sql files and compare expected results. 🙏🏼

crates/pg_analyser/CONTRIBUTING.md
juleswritescode: Ha, this is actually makes sense, because there are triple backticks *within* the code block. 👍

psteinroe (approved):
Thanks!

PR #196, opened 2025-02-09
chore: rename crates to pglt

psteinroe:
... that's all 😅 i added the config to rename the cli binary in  a follow up pr


juleswritescode (approved):
I didn't check every line but LGTM :)

PR #199, opened 2025-02-11
fix(pglt_statement_splitter): alter column split

psteinroe:
fixes splitter for `alter table alter column` statements and re-enables check within rules diagnostics. we did not ignore `ALTER` within an `ALTER` statement before.


juleswritescode (approved):
Epic! Easy change, this means the splitter is well written :)

PR #203, opened 2025-02-12
test(analyser): generate tests with `new-lintrule` script and add testing suite

juleswritescode:
The goal is to have something similar to Biome's [testing suite](https://github.com/biomejs/biome/blob/main/crates/biome_css_analyze/tests/spec_tests.rs).


juleswritescode:
Will add a Readme tomorrow, but it's reviewable 👍

psteinroe (approved):
nice!

PR #204, opened 2025-02-13
chore(docs): initial commit

psteinroe:
adds initial docs with generated documentation wherever possible.
- [x] codegen for default config
- [x] codegen for env vars
- [x] codegen for cli reference
- [x] codegen for lint rules
- [x] codegen for config schema, but I think we dont use schemars everywhere already. I checked and it's possible to use json schemas for toml files so we should at least add it at some point. (done in #206)
- [x] Integrate docs codegen in CI

also renamed a few leftovers to PGLT

Next, we need to focus on the specific use cases and provide guides for them, especially "lint your migration files".

### DIFF
--- a/docs/codegen/src/rules_docs.rs
+++ b/docs/codegen/src/rules_docs.rs
@@ -91,6 +91,7 @@ fn generate_rule_doc(
 
     writeln!(content)?;
     writeln!(content, "**Since**: `v{}`", meta.version)?;
+    writeln!(content)?;
 
     // add recommended notice
     if is_recommended {
@@ -105,7 +106,7 @@ fn generate_rule_doc(
 
     // add source information
     if !meta.sources.is_empty() {
-        writeln!(content, "Sources: ")?;
+        writeln!(content, "**Sources**: ")?;
 
         for source in meta.sources {
             let rule_name = source.to_namespaced_rule_name();
--- a/docs/codegen/src/rules_docs.rs
+++ b/docs/codegen/src/rules_docs.rs
@@ -1,0 +1,469 @@
+use anyhow::{bail, Result};
+use biome_string_case::Case;
+use pglt_analyse::{AnalyserOptions, AnalysisFilter, RuleFilter, RuleMetadata};
+use pglt_analyser::{Analyser, AnalyserConfig};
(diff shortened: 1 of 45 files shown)


docs/codegen/src/rules_docs.rs
```diff
@@ -0,0 +1,464 @@
...
+            write!(content, "- Inspired from: ")?;
+            writeln!(
+                content,
+                "<a href=\"{source_rule_url}\" target=\"_blank\"><code>{rule_name}</code></a>"
+            )?;
+        }
+        writeln!(content)?;
+    }
+
+    write_documentation(group, rule, meta.docs, &mut content)?;
```
juleswritescode, line 100: I just realized that the `declare_lint_rule` macro will use doc comment and attach it as a `.doc` property. I'm impressed 🤯

docs/codegen/src/rules_docs.rs
```diff
@@ -0,0 +1,464 @@
...
+}
+
+impl FromStr for CodeBlockTest {
+    type Err = anyhow::Error;
+
+    fn from_str(input: &str) -> Result<Self> {
+        // This is based on the parsing logic for code block languages in `rustdoc`:
+        // https://github.com/rust-lang/rust/blob/6ac8adad1f7d733b5b97d1df4e7f96e73a46db42/src/librustdoc/html/markdown.rs#L873
+        let tokens = input
+            .split([',', ' ', '\t'])
```
juleswritescode, line 369: rust is such a great language

docs/codegen/src/utils.rs
```diff
@@ -0,0 +1,55 @@
...
+    re.replace_all(content, format!("${{1}}{}${{2}}", replacement))
+        .to_string()
+}
+
+#[derive(Default)]
+pub(crate) struct LintRulesVisitor {
+    /// This is mapped to:
+    /// - group (correctness) -> list of rules
+    /// - list or rules is mapped to
+    /// - rule name -> metadata
```
juleswritescode, line 24:
```suggestion
    /// group (e.g. "safety") -> <list of rules>
    /// where <list of rules> is:
    /// <rule name> -> metadata
```

juleswritescode (approved):
Looks great to me :)

PR #206, opened 2025-02-17
feat(docs): generate json schema for config

psteinroe:
fixes schemars integration in `pglt_configuration` and adds a little codegen to generate the schemas. we safe them in `docs/schemas/latest/schema.json` and `docs/schemas/VERSION/schema.json`. I also added the latest `schema.json` to the release assets.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,7 +63,7 @@ jobs:
       - name: 🛠️ Run Build
         run: cargo build -p pglt_cli --release --target ${{ matrix.config.target }}
 
-      # windows is a special snowflake to, it saves binaries as .exe
+      # windows is a special snowflake too, it saves binaries as .exe
       - name: 👦 Name the Binary
         if: matrix.config.os == 'windows-latest'
         run: |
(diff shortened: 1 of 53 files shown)


.github/workflows/release.yml
```diff
@@ -124,7 +124,9 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           body: ${{ steps.create_changelog.outputs.content }}
           tag_name: ${{ steps.create_changelog.outputs.version }}
-          files: pglt_*
+          files: |
+            pglt_*
+            docs/schemas/latest/schema.json
```
juleswritescode, line 129: Nice! Should we also regenerate the schema as part of the release workflow?

juleswritescode (changes requested):

.github/workflows/release.yml
psteinroe: we already check that the schema is generated as part of the pull request workflow, so it should always be up to date if changes are made via pull requests

juleswritescode (approved):

PR #208, opened 2025-02-18
feat(rules): adding-required-field

juleswritescode:


juleswritescode:
@psteinroe The codegen CI keeps failing, I'm not sure if that is my mistake? 
I expected it to generate the docs automatically on PR, but it fails gecause of the `git diff` command, even though the diff seems the same:

Screenshot 2025-02-19 at 09 30 24

Should I run the command manually before I push?

psteinroe:
ah yes, right now we don't generate + commit automatically. I was thinking to add a "just ready" command to run before opening a pr that does all of this?

juleswritescode:
> ah yes, right now we don't generate + commit automatically. I was thinking to add a "just ready" command to run before opening a pr that does all of this?

That would help! And maybe we could show a suggestion in the workflow that tells the developer to run the command?

crates/pglt_analyser/src/lint/safety/adding_required_field.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// This will fail immediately upon running for any populated table. Furthermore, old application code that is unaware of this column will fail to INSERT to this table.
+    ///
+    /// Make new columns optional initially by omitting the NOT NULL constraint until all existing data and application code has been updated. Once no NULL values are written to or persisted in the database, set it to NOT NULL.
+    /// Alternatively, if using Postgres version 11 or later, add a DEFAULT value that is not volatile. This allows the column to keep its NOT NULL constraint.
+    ///
+    /// ## Invalid
+    /// alter table test add column count int not null;
+    ///
+    /// ## Valid in Postgres >= 11
+    /// alter table test add column count int not null default 0;
```
psteinroe, line 16: we should pass the current postgres version to the rules!

psteinroe (approved):
lgtm! but we should think about integrating the current pg version into the analyser

PR #209, opened 2025-02-18
fix: check for lint fix in workflow

psteinroe:
runs `lint --fix` in the workflow and checks for any diffs. this should reduce unrelated changes in prs.


juleswritescode (approved):
I like it!

We could alternatively run this pre-push or pre-commit; we do this at Parqet and I like it (some people don't though).

psteinroe:
> I like it!
> 
> We could alternatively run this pre-push or pre-commit; we do this at Parqet and I like it (some people don't though).

I would prefer to keep that decision on the developer

PR #213, opened 2025-02-21
fix(pglt_workspace): lock

psteinroe:
fixes our bug from last night. the issue was actually that the entire server went into a lock when loading the `SchemaCache` for auto-completion, because we did not release the read lock on `inner` before acquiring the write lock. I wonder how we could write a test for these kind of issues 🤔 maybe some kind of simulation e2e test where we actually start the server and start to send some commands and then stop it again. 

also removed some spammy logs.

### DIFF
--- a/crates/pglt_workspace/src/workspace/server/schema_cache_manager.rs
+++ b/crates/pglt_workspace/src/workspace/server/schema_cache_manager.rs
@@ -42,23 +42,25 @@ pub struct SchemaCacheManager {
 
 impl SchemaCacheManager {
     pub fn load(&self, pool: PgPool) -> Result<SchemaCacheHandle, WorkspaceError> {
-        let inner = self.inner.read().unwrap();
+        let new_conn_str = pool_to_conn_str(&pool);
 
-        if pool_to_conn_str(&pool) == inner.conn_str {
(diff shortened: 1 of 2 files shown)


crates/pglt_workspace/src/workspace/server/schema_cache_manager.rs
```diff
@@ -42,23 +42,25 @@ pub struct SchemaCacheManager {
...
-            let new_conn_str = pool_to_conn_str(&pool);
+        let new_conn_str = pool_to_conn_str(&pool);
+
+        {
+            // return early if the connection string is the same
+            let inner = self.inner.read().unwrap();
+            if new_conn_str == inner.conn_str {
+                return Ok(SchemaCacheHandle::wrap(inner));
+            }
+        }
```
psteinroe, line 53: we wrap this into its own block so the lock is dropped after

juleswritescode (approved):
Thanks for cleaning that up!

psteinroe:
> Thanks for cleaning that up! 
> 
> 
> 
> We could add a simple regression test to the schema cache manager?

Yeah in this case it should actually be sufficient to just call ist once 😂

PR #214, opened 2025-02-21
chore(pglt_lsp): add lifecycle test

psteinroe:
adds a basic lifecycle test to the lsp server. will expand on it later. maybe we can even have a "recording" of the users actions via logs, parse these recordings and replay them.

### DIFF
--- a/crates/pglt_lsp/tests/server.rs
+++ b/crates/pglt_lsp/tests/server.rs
@@ -1,0 +1,341 @@
+use anyhow::bail;
+use anyhow::Context;
+use anyhow::Error;
+use anyhow::Result;
+use futures::channel::mpsc::{channel, Sender};
+use futures::Sink;
+use futures::SinkExt;
+use futures::Stream;
+use futures::StreamExt;
+use pglt_lsp::LSPServer;
+use pglt_lsp::ServerFactory;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+use serde_json::{from_value, to_value};
+use std::any::type_name;
+use std::fmt::Display;
+use std::time::Duration;
+use tower::timeout::Timeout;
+use tower::{Service, ServiceExt};
+use tower_lsp::jsonrpc;
+use tower_lsp::jsonrpc::Response;
+use tower_lsp::lsp_types as lsp;
+use tower_lsp::lsp_types::DidCloseTextDocumentParams;
+use tower_lsp::lsp_types::DidOpenTextDocumentParams;
+use tower_lsp::lsp_types::InitializeResult;
+use tower_lsp::lsp_types::InitializedParams;
+use tower_lsp::lsp_types::PublishDiagnosticsParams;
+use tower_lsp::lsp_types::TextDocumentContentChangeEvent;
+use tower_lsp::lsp_types::TextDocumentIdentifier;
+use tower_lsp::lsp_types::TextDocumentItem;
+use tower_lsp::lsp_types::VersionedTextDocumentIdentifier;
+use tower_lsp::lsp_types::{ClientCapabilities, Url};
+use tower_lsp::lsp_types::{DidChangeConfigurationParams, DidChangeTextDocumentParams};
+use tower_lsp::LspService;
+use tower_lsp::{jsonrpc::Request, lsp_types::InitializeParams};
+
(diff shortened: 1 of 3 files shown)


crates/pglt_lsp/tests/server.rs
```diff
@@ -0,0 +1,341 @@
...
+use tower_lsp::lsp_types as lsp;
+use tower_lsp::lsp_types::DidCloseTextDocumentParams;
+use tower_lsp::lsp_types::DidOpenTextDocumentParams;
+use tower_lsp::lsp_types::InitializeResult;
+use tower_lsp::lsp_types::InitializedParams;
+use tower_lsp::lsp_types::PublishDiagnosticsParams;
+use tower_lsp::lsp_types::TextDocumentContentChangeEvent;
+use tower_lsp::lsp_types::TextDocumentIdentifier;
+use tower_lsp::lsp_types::TextDocumentItem;
+use tower_lsp::lsp_types::VersionedTextDocumentIdentifier;
```
juleswritescode, line 31: the use statements look a bit odd? something wrong with your nvim config?

crates/pglt_lsp/tests/server.rs
```diff
@@ -0,0 +1,341 @@
...
+/// doesn't have to refer to an existing file on the host machine.
+macro_rules! url {
+    ($path:literal) => {
+        if cfg!(windows) {
+            lsp::Url::parse(concat!("file:///z%3A/workspace/", $path)).unwrap()
+        } else {
+            lsp::Url::parse(concat!("file:///workspace/", $path)).unwrap()
+        }
+    };
+}
```
juleswritescode, line 49: pretty cool

crates/pglt_lsp/tests/server.rs
```diff
@@ -0,0 +1,341 @@
...
+                Request::build(method)
+                    .params(to_value(&params).context("failed to serialize params")?)
+                    .finish(),
+            )
+            .await
+            .map_err(Error::msg)
+            .context("call() returned an error")
+            .and_then(|res| {
+                if let Some(res) = res {
+                    bail!("shutdown returned {:?}", res)
```
juleswritescode, line 81: Are we shutting down here?

crates/pglt_lsp/tests/server.rs
```diff
@@ -0,0 +1,341 @@
...
+            .context("ready() returned an error")?
+            .call(
+                Request::build(method)
+                    .params(to_value(&params).context("failed to serialize params")?)
+                    .finish(),
+            )
+            .await
+            .map_err(Error::msg)
+            .context("call() returned an error")
+            .and_then(|res| {
```
juleswritescode, line 79: Hmm, would it be more flexible to let test callers handle the response themselves?

crates/pglt_lsp/tests/server.rs
```diff
@@ -0,0 +1,341 @@
...
+            "textDocument/didOpen",
+            DidOpenTextDocumentParams {
+                text_document: TextDocumentItem {
+                    uri: url!("document.js"),
+                    language_id: String::from("javascript"),
+                    version: 0,
+                    text: text.to_string(),
+                },
+            },
+        )
```
juleswritescode, line 190: Just a feeling: I think we need to be careful about what we abstract away and what we force the tests to do themselves.

For methods that are always called like `shutdown`, `initialize` etc. I think it makes sense to provide convenience methods.

But I'm not super about e.g. `open_document` – a bug might slip in because our testing logic contains too much "business logic". 

Also, this is javascript 🤣 

But yeah, WDYT? I think it's a fine line

juleswritescode (approved):
Just suggestions/nits 👍👍

crates/pglt_lsp/tests/server.rs
psteinroe: copy-pasta 😂 fixed it

crates/pglt_lsp/tests/server.rs
psteinroe: 9l0d7o

crates/pglt_lsp/tests/server.rs
psteinroe: yeah but every call will want to deserialise it I guess. but lets reiterate when we wrote a few tests 👍

PR #217, opened 2025-02-21
chore: add publish npm action

juleswritescode:
The publish action will be a bit different from biome's because we want to review our Releases as Drafts.

next steps:

- [x] replicate all the relevant /packages/@pglt/* folders
- [x] verify that downloading assets works
- [x] make sure we can publish prelease versions
- [x] once all that's done, we can publish to NPM

then head over to publishing for brew

### DIFF
--- a/packages/@pglt/pglt/package.json
+++ b/packages/@pglt/pglt/package.json
@@ -33,8 +33,8 @@
     "provenance": true
   },
   "optionalDependencies": {
-    "pglt-x86_64-windows-msvc.exe": "<placeholder>",
-    "pglt-aarch64-windows-msvc.exe": "<placeholder>",
+    "pglt-x86_64-windows-msvc": "<placeholder>",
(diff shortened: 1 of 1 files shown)


psteinroe (approved):
great work! should we already update it to the correct names before merging or merge it now or keep it open until we have the input from supabase?

juleswritescode:
@psteinroe this is not the most up-to-date version, I'll push that from my private branch later 👍 
Regarding the name: I don't mind, we can also deploy to a different package name and delete it later. Shouldn't be a problem right now, I did that a couple of times on npm 👍

juleswritescode:
@psteinroe Okay it's done! The package is now named `pglt` and the subpackages e.g. `pglt-aarch64-linux`. 
It would be cool if you could double-check all the names and repository URLS – already it it twice, but I'm a bit tired 
🤓

PR #218, opened 2025-02-22
chore(pglt_lsp): more tests

psteinroe:
Adds two more lifecycle test. One with a db connection to ensure we don't deadlock, and one that runs the custom shutdown command. Also moved `insta` to the workspace level.


juleswritescode (approved):
Nice!!

PR #219, opened 2025-02-22
fix(pglt_workspace): emit scan errors

psteinroe:
emitting scan errors from the lexer. we ignored them before because I thought they never happen. but they do! 

- added specialised diagnostic to both the lexer and the statement splitter 
- bubble up error from lexer 
- store parse diagnostics on document 
- if any diagnostics are fatal, we remove all statements and only process text changes until they are gone
- expose document diagnostics 

also:
- removed the leftover base_db crate


crates/pglt_lexer/src/diagnostics.rs
```diff
@@ -0,0 +1,63 @@
+use pglt_diagnostics::{Diagnostic, MessageAndDescription};
+use text_size::TextRange;
+
+/// A specialized diagnostic for scan errors.
+///
+/// Scan diagnostics are always **fatal errors**.
+#[derive(Clone, Debug, Diagnostic, PartialEq)]
+#[diagnostic(category = "syntax", severity = Fatal)]
```
psteinroe, line 8: `severity = Fatal` is important here -> its what we use higher up to determine whether the document can be used or not

crates/pglt_lexer/src/diagnostics.rs
```diff
@@ -0,0 +1,63 @@
...
+                    let to = from + search_term.len();
+                    diagnostics.push(ScanError {
+                        span: Some(TextRange::new(
+                            from.try_into().unwrap(),
+                            to.try_into().unwrap(),
+                        )),
+                        message: MessageAndDescription::from(err_msg.clone()),
+                    });
+                }
+            }
```
psteinroe, line 38: we try to find the problematic text in the input to put a range on the diagnostics. this is not safe in any way but I find it nicer to show too many diagnostics where one is for sure valid than none and have the user look through the text.

crates/pglt_lexer/src/diagnostics.rs
```diff
@@ -0,0 +1,63 @@
...
+impl ScanError {
+    pub fn from_pg_query_err(err: pg_query::Error, input: &str) -> Vec<Self> {
+        let err_msg = err.to_string();
+        let re = regex::Regex::new(r#"at or near "(.*?)""#).unwrap();
+        let mut diagnostics = Vec::new();
+
+        for captures in re.captures_iter(&err_msg) {
+            if let Some(matched) = captures.get(1) {
+                let search_term = matched.as_str();
+                for (idx, _) in input.match_indices(search_term) {
```
juleswritescode, line 27: what happens if we have multiple matches in an input? Is that possible?

crates/pglt_lexer/src/diagnostics.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::lex;
+
+    #[test]
+    fn finds_all_occurrences() {
+        let input =
+            "select 1443ddwwd33djwdkjw13331333333333; select 1443ddwwd33djwdkjw13331333333333;";
```
juleswritescode, line 59: the diagnostics should have distinct ranges, should we assert that as well?

crates/pglt_workspace/src/workspace/server/change.rs
```diff
@@ -235,7 +259,22 @@ impl Document {
...
-            let new_ranges = pglt_statement_splitter::split(changed_content).ranges;
+            let (new_ranges, diags) =
+                document::split_with_diagnostics(changed_content, Some(affected_range.start()));
+
+            self.diagnostics = diags;
+
+            if self
+                .diagnostics
+                .iter()
+                .any(|d| d.severity() == pglt_diagnostics::Severity::Fatal)
```
juleswritescode, line 270: hard to see in the github diff, but is `self` not a `Document` and we can check the fatal_error helper?

crates/pglt_workspace/src/workspace/server/change.rs
```diff
@@ -289,7 +328,22 @@ impl Document {
...
-        let new_ranges = pglt_statement_splitter::split(changed_content).ranges;
+        let (new_ranges, diags) =
+            document::split_with_diagnostics(changed_content, Some(full_affected_range.start()));
+
+        self.diagnostics = diags;
+
+        if self
+            .diagnostics
+            .iter()
+            .any(|d| d.severity() == pglt_diagnostics::Severity::Fatal)
```
juleswritescode, line 339: hard to see in the github diff, but is `self` not a `Document` and we can check the fatal_error helper?

crates/pglt_workspace/src/workspace/server/change.rs
```diff
@@ -69,37 +75,55 @@ impl Document {
...
-                    })
-                }),
-        );
+        self.diagnostics = diagnostics;
+
+        // Do not add any statements if there is a fatal error
+        if self
+            .diagnostics
+            .iter()
+            .any(|d| d.severity() == pglt_diagnostics::Severity::Fatal)
```
juleswritescode, line 109: hard to see in the github diff, but is `self` not a `Document` and we can check the fatal_error helper?

crates/pglt_workspace/src/workspace/server/change.rs
```diff
@@ -289,7 +328,22 @@ impl Document {
...
+        {
+            // cleanup all positions if there is a fatal error
+            changed.extend(self.drain_positions());
+            // still process text change
+            self.content = new_content;
+            return changed;
```
juleswritescode, line 345: nit: duplicate logic

juleswritescode (approved):
Nice work :)

crates/pglt_lexer/src/diagnostics.rs
psteinroe: Yes, we return two diagnostics. This is what I meant with my comment about possibly incorrect diagnostics.

psteinroe: ahhh okay mh yes that's correct but I don't think it's an issue since this is coming from Postgres and even if there are multiple occurrences the first will be the correct one.

PR #220, opened 2025-03-01
docs: add short guide for migrations

psteinroe:
We should also add a section for --since and --changes


juleswritescode (approved):
Well written 👍

PR #224, opened 2025-03-07
chore: rust 2024

psteinroe:
updated to rust 2024. all changes are from
- `cargo fix --edition`
- `just lint-fix`
- `just format`

also had to update the `version` of `rustfmt` and make sure that `xtask` format also uses `nightly`.

the GitHub action lint is fixed in #216


juleswritescode (approved):
swee-heet

PR #230, opened 2025-03-11
fix: whitespace-only change 

psteinroe:
fixes an issue where a whitespace-only change within a single statement was not chaining the range of the affected statement, only the ranges of the statements that are after the end of the affected one.

### DIFF
--- a/crates/pglt_workspace/src/workspace/server/change.rs
+++ b/crates/pglt_workspace/src/workspace/server/change.rs
@@ -268,18 +268,6 @@ impl Document {
             }
 
             if new_ranges.len() == 1 {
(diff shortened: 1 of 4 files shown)


crates/pglt_workspace/src/workspace/server/change.rs
```diff
@@ -648,6 +643,83 @@ mod tests {
         assert_document_integrity(&d);
     }
 
+    #[test]
+    fn within_statements_2() {
```
psteinroe, line 647: this test was failing before my fix because of document integrity

juleswritescode (approved):

PR #231, opened 2025-03-11
fix: another whitespace change issue 

psteinroe:
we need to make sure that our change_range is always valid


juleswritescode (approved):
makes sense!

PR #232, opened 2025-03-12
fix: emit whitespace changes after all

psteinroe:
after some forth and back yesterday and a crash when starting to complete after a change I realised I am big idiot because not emitting whitespaces will result in an invalid tree sitter tree… 🥲 so yeah, we just emit them now. also added an e2e test for completions.

also
- removed changed statements because we do not use them and they will jus grow over time 
- my format-on-write did minor changes to the sql in tests


juleswritescode (approved):
I highly doubt that you're a big idiot, you're not even 1.90m

PR #238, opened 2025-03-14
feat: pglt vscode extension

juleswritescode:
I'm not sure if we should keep the extension in this repository. 
[other libraries](https://github.com/biomejs/biome-vscode) keep theirs in a separate repo. 
I think it'll need some maintainence work and a separate changelog, tbh!


juleswritescode:
@psteinroe just created the PR so you can get an overview. WDYT, separate repo?

PR #239, opened 2025-03-14
refactor: migrate to jsonc config

psteinroe:
as per title. 

our new config is `.jsonc`, which is "json with comments". serde doesnt support comments so we strip them before parsing. 

also unified all `rename_all` configs to `camelCase`.

and added `$schema` to the `jsonc` config so we should get auto-complete if the editor supports it. 

we will need to update the path to the schema file (if its not available locally, which is preferred) once we have our domain.


crates/pglt_workspace/src/configuration.rs
```diff
@@ -168,7 +169,20 @@ pub fn create_config(
         }
     })?;
 
-    let contents = toml::ser::to_string_pretty(&configuration)
+    // we now check if biome is installed inside `node_modules` and if so, we use the schema from there
+    if VERSION == "0.0.0" {
+        let schema_path = Path::new("./node_modules/@pglt/pglt/schema.json");
```
juleswritescode, line 174: den Part verstehe ich nicht so ganz, die früheste Version auf NPM wird ja >0.1.0 sein?

crates/pglt_workspace/src/configuration.rs
```diff
@@ -186,3 +200,68 @@ pub fn to_analyser_rules(settings: &Settings) -> AnalyserRules {
     }
     analyser_rules
 }
+
+/// Takes a string of jsonc content and returns a comment free version
+/// which should parse fine as regular json.
+/// Nested block comments are supported.
+pub fn strip_jsonc_comments(jsonc_input: &str) -> String {
```
juleswritescode, line 207: wollen wir hierfür noch ein paar unit tests hinzufügen?

crates/pglt_workspace/src/configuration.rs
```diff
@@ -168,7 +169,20 @@ pub fn create_config(
         }
     })?;
 
-    let contents = toml::ser::to_string_pretty(&configuration)
+    // we now check if biome is installed inside `node_modules` and if so, we use the schema from there
```
juleswritescode, line 172:
```suggestion
    // we now check if pglt is installed inside `node_modules` and if so, we use the schema from there
```

crates/pglt_workspace/src/configuration.rs
psteinroe: done

juleswritescode (approved):

PR #241, opened 2025-03-18
chore: rename pglt in packages/

juleswritescode:
simple find and replace in packages/.

renamed these to `pgt/**`:

Screenshot 2025-03-18 at 08 19 54


psteinroe:
Nice! Yes, we should rename them to pgt. Also the PgLTPath should be PgTPath

psteinroe:
Note that the tests fail now because we will need to rename stuff in the crates at the same time. Will do so later

PR #255, opened 2025-03-20
fix: various fixes across the board

psteinroe:
- switch back to async-std for sqlx
- fix remaining renamings and urls


crates/pgt_lsp/tests/server.rs
```diff
@@ -441,7 +441,7 @@ async fn server_shutdown() -> Result<()> {
     let cancellation = factory.cancellation();
     let cancellation = cancellation.notified();
 
-    // this is called when `pglt stop` is run by the user
+    // this is called when `pgt stop` is run by the user
```
juleswritescode, line 444:
```suggestion
    // this is called when `postgrestools stop` is run by the user
```

juleswritescode (approved):

PR #257, opened 2025-03-22
feat: execute stmt under cursor

### LINKED ISSUE #191: Bring back the commands crate
# Planned Work
re-integrate the `commands` feature crate, expose it in the workspace api and integrate it into the language server.

We removed it when migrating to the new infrastructure.

juleswritescode:
closes #191


crates/pgt_lsp/src/handlers/code_actions.rs
```diff
@@ -0,0 +1,114 @@
...
+                    statement_id: id,
+                    path,
+                })?;
+
+            /**
+             * Updating all diagnostics: the changes caused by the statement execution
+             * might affect many files.
+             *
+             * TODO: in test.sql, this seems to work after create table, but not after drop table.
+             */
```
psteinroe, line 101: dont we need to invalidate the schema cache for this?

crates/pgt_workspace/src/code_actions.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub struct CodeAction {
+    pub title: String,
+    pub kind: CodeActionKind,
+    pub disabled_reason: Option<String>,
+}
+
```
psteinroe, line 28: maybe we can put this in to a `features/` directory? we could also structure the other feature params and types in a similar manner.

crates/pgt_workspace/src/workspace/server.rs
```diff
@@ -253,6 +259,88 @@ impl Workspace for WorkspaceServer {
...
+    ) -> Result<code_actions::CodeActionsResult, WorkspaceError> {
+        let doc = self
+            .documents
+            .get(&params.path)
+            .ok_or(WorkspaceError::not_found())?;
+
+        let eligible_statements = doc
+            .iter_statements_with_text_and_range()
+            .filter(|(_, range, _)| range.contains(params.cursor_position));
+
```
psteinroe, line 274: we should check that the statement is valid by checking if there is a pg_query AST for it

psteinroe (changes requested):
most of it are nits, but I think we should check if the statement is valid before offering the execution

crates/pgt_workspace/src/workspace/server.rs
juleswritescode: great point, done.

crates/pgt_lsp/src/handlers/code_actions.rs
juleswritescode: As far as I can tell, we're not using the SchemaCache for diagnostics, we're just using a PREPARE statement against the connection

crates/pgt_lsp/src/handlers/helper.rs
```diff
@@ -0,0 +1,26 @@
+use crate::session::Session;
+use pgt_text_size::TextSize;
+use tower_lsp::lsp_types;
+
+pub fn get_cursor_position(
```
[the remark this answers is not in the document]
juleswritescode: I think that makes sense! Are you good if we do this in a separate PR as well?

psteinroe (approved):
awesome! 🙌

PR #266, opened 2025-03-29
chore: improve logs

juleswritescode:
1. Adds a configurable `bunyan` log format (default is still hierarchical), configurable via PGT_LOG_KIND env var
2. Adds timestamps to hierarchical log format
3. Makes log level configurable via PGT_LOG_LEVEL env var, default stays the same
4. Cleans up log journey by removing logged fields that we do not use at runtime


juleswritescode:
@psteinroe The default log level is now `info` – if we ask issue reporters to copy their log files, maybe the default should be `debug`. wdyt?

psteinroe (approved):
thanks for the flexibility!

Issue #174, opened 2025-01-28
add wasm integration


psteinroe:
we want to make the platform as accessible as possible, and a wasm integration would allow us to run the workspace api within a service worker. this would allow tools like [database.build](https://database.build) or other [pglite](https://pglite.dev)-based solutions to offer language features right in the browser.

we can use [`biome_wasm`](https://github.com/biomejs/biome/tree/main/crates/biome_wasm) as an inspiration, and also need to publish npm packages.

psteinroe:
Update: this is blocked by `pg_query.rs` which doesn't build to wasm yet.

PR #268, opened 2025-03-29
feat: put db con behind feature gate

psteinroe:

puts database connection behind a feature gate `db-connection`. the flag is enabled by default, so nothing changes. this is premiliary work for #174. after this pr, my plan is to
- provide a wasm build without any database first and then 
- dive deep into integrating pglite. right now, I believe there is still work to be done on their side for us to make the integration seamless.

sqlx and database features are built all the time, leading to a failing build for wasm.

## What is the new behavior?

we can exclude sqlx from the build by disabling the default feature flag `db-connection`


crates/pgt_workspace/src/workspace.rs
```diff
@@ -53,6 +53,73 @@ pub struct GetCompletionsParams {
...
+    Table,
+    Function,
+    Column,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub struct CompletionItem {
+    pub label: String,
+    pub(crate) score: i32,
```
psteinroe, line 117: question: do we even need the score here?

PR #272, opened 2025-03-29
regrouping installation options

kiwicopple:
Feel free to edit this. The goal is to create a clearer logical flow:

- overview
- install
- configure
- use

previously the installation instructions are a bit scattered


psteinroe (approved):
thanks for jumping in!

PR #277, opened 2025-03-30
fix: clean `cargo check` and `cargo clippy` 

juleswritescode:
Aiming for a clean `cargo check` and `cargo clippy`. 

I tried to remove only what we'll probably not need (e.g. `ClientInformation`, since we'll only handle 1 client per server) and what we can easily replace (e.g. enum-check-convenience methods).

There's still a tiny tree-sitter warning from `scanner.c` :)


psteinroe (approved):
noice

PR #291, opened 2025-04-01
fix: execute statement reliable

juleswritescode:
I've been through many iterations but this seems to be the simplest.

We use a dedicated connection for the `prepare` statement and close it on drop. This will clear all the prepared statements for that session. `sqlx` itself doesn't cache anything when using `prepare`. If we worry about performance overhead, I suggest we a) use sqlx cache via `prepare_with` and b) debounce the publishing of diagnostics again 👍

### DIFF
--- a/crates/pgt_lsp/src/handlers/code_actions.rs
+++ b/crates/pgt_lsp/src/handlers/code_actions.rs
@@ -1,3 +1,5 @@
+use std::sync::atomic::Ordering;
+
 use crate::{adapters::get_cursor_position, session::Session};
 use anyhow::{Result, anyhow};
 use tower_lsp::lsp_types::{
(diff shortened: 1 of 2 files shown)


crates/pgt_lsp/src/handlers/code_actions.rs
```diff
@@ -1,3 +1,5 @@
+use std::sync::atomic::Ordering;
```
psteinroe, line 1: leftover I think?

psteinroe (approved):
after the ordering was removed this makes me very happy! there is one unused import but other than that great job!

crates/pgt_lsp/src/handlers/code_actions.rs
juleswritescode: yes

PR #301, opened 2025-04-04
fix: support non timestamp but numeric migrations

### LINKED ISSUE #300: add support for non-timestamp but still numeric migrations
### Discussed in https://github.com/supabase-community/postgres-language-server/discussions/298

<sup>Originally posted by **mplanchard** April  3, 2025</sup>
Right now, it seems like the `after` configuration for migrations is closely coupled with a file format used by certain ORMs (e.g. prisma).

In our project, the files are named like `0001_whatever.sql`, with no timestamp.

If `after` is going to refer to a timestamp rather than a migration number, it would be nice to do it in a more generalized way. The first thing that comes to mind is using git to determine the file's last modification time, if VCS integration is enabled.

Alternatively, a configuration option to use migration numbers would be nice, which could potentially take some kind of DSL to convert filenames to migration numbers (perhaps a sed command).</div>

psteinroe:
fixes #300

### DIFF
--- a/crates/pgt_workspace/src/workspace/server.rs
+++ b/crates/pgt_workspace/src/workspace/server.rs
@@ -108,7 +108,7 @@ impl WorkspaceServer {
                 let migrations_dir = migration_settings.path.as_ref()?;
                 let migration = migration::get_migration(path, migrations_dir)?;
 
-                Some(&migration.timestamp <= ignore_before)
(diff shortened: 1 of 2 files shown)


juleswritescode (approved):
LGTM, but: Should we rename the `timestamp` property, since it's more of an ID now?

PR #306, opened 2025-04-04
fix: undetermined string error 

psteinroe:
also cannot repro manually:
Screenshot 2025-04-04 at 11 33 01

### DIFF
--- a/crates/pgt_lsp/tests/server.rs
+++ b/crates/pgt_lsp/tests/server.rs
@@ -981,3 +981,89 @@ async fn test_issue_281() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_issue_303() -> Result<()> {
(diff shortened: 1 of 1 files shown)


crates/pgt_lsp/tests/server.rs
```diff
@@ -981,3 +981,89 @@ async fn test_issue_281() -> Result<()> {
...
+
+    server.initialize().await?;
+    server.initialized().await?;
+
+    server.load_configuration().await?;
+
+    server.open_document("").await?;
+
+    let chars = [
+        "C", "R", "E", "A", "T", "E", " ", "T", "A", "B", "L", "E", " ", "\"",
```
juleswritescode, line 1040: sanity check, do these need to be of `char` type? `'C', '"'` etc.

psteinroe:
fixed by moving to a custom lexer

PR #310, opened 2025-04-04
feat: request autocompletion without typing a letter

juleswritescode:
So far, we were relying on complete tree-sitter trees for autocompletion.

This works fine if you type the first letter of a node, say:
`select * from u|` 
Now, the `u` is considered a table node, and we can look for tables.

But this PR makes it such that we can also request completions if we haven't typed the `u` yet.

It does so by checking whether the tree is valid and, if not, injecting a token to fix the tree-sitter tree.

Additionally, in the case of `select * from |`, the workspace would consider the cursor to be outside the statement. I added a little offset so the cursor still matches a statement if it's 2 characters after the last token.

### DIFF
--- a/crates/pgt_completions/src/sanitization.rs
+++ b/crates/pgt_completions/src/sanitization.rs
@@ -16,20 +16,18 @@ pub fn benchmark_sanitization(params: CompletionParams) -> String {
     params.text
 }
 
-impl<'larger, 'smaller> TryFrom<CompletionParams<'larger>> for SanitizedCompletionParams<'smaller>
+impl<'larger, 'smaller> From<CompletionParams<'larger>> for SanitizedCompletionParams<'smaller>
 where
     'larger: 'smaller,
 {
-    type Error = String;
-
-    fn try_from(params: CompletionParams<'larger>) -> Result<Self, Self::Error> {
+    fn from(params: CompletionParams<'larger>) -> Self {
         if cursor_inbetween_nodes(params.tree, params.position)
             || cursor_prepared_to_write_token_after_last_node(params.tree, params.position)
             || cursor_before_semicolon(params.tree, params.position)
         {
-            Ok(SanitizedCompletionParams::with_adjusted_sql(params))
+            SanitizedCompletionParams::with_adjusted_sql(params)
         } else {
-            Ok(SanitizedCompletionParams::unadjusted(params))
+            SanitizedCompletionParams::unadjusted(params)
         }
     }
 }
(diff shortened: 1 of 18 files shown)


crates/pgt_completions/src/sanitization.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        for (idx, c) in params.text.chars().enumerate() {
+            if idx == cursor_pos {
+                sql.push_str(SANITIZED_TOKEN);
+                sql.push(' ');
+            }
+            sql.push(c);
+        }
+
+        let mut parser = tree_sitter::Parser::new();
```
psteinroe, line 55: isnt there a way to modify the tree sitter tree instead of reparsing the string?

psteinroe (commented):
looks good to me overall. just would like to double check if we really need to instantiate the tree sitter parser with every request 😢. can't we just mutate the tree?

also: we will have a bunch of conflicts, and a few things will need to be handled differently. but no issues, just that we should merge yours first so I can fix the merge conflicts with mine.

crates/pgt_completions/src/sanitization.rs
psteinroe: I think editing the node isn to possible, so maybe we should try to pass the existing tree sitter parser down.

juleswritescode:
> looks good to me overall. just would like to double check if we really need to instantiate the tree sitter parser with every request 😢. can't we just mutate the tree?

juleswritescode:
@psteinroe Curious about your opinion: I tried moving the `GetCompletionsMapper` into the completions crate, but then I couldn't access the `parser.cst_db` because that's private. 

We could make the `parser.cst_db` public, or provide a `parser.get_tree()` method, or leave the mappers in the crate… what do you think?

psteinroe:
ah sorry, i didn't mean to move the mapper there. I was thinking to clone it copy the tree sitter parser to skip the initialisation costs. 

Will look at the pr later, thanks for trying! 🙏🏼

crates/pgt_workspace/src/features/completions.rs
```diff
@@ -24,3 +28,167 @@ impl IntoIterator for CompletionsResult {
         self.items.into_iter()
     }
 }
+
+pub(crate) fn get_statement_for_completions<'a>(
```
psteinroe, line 32: why can't this be implemented as a `Filter` in `parsed_document`?

psteinroe (approved):
just one detail, rest lgtm! I would say we go with constructing a new parser with every request for now and then check it later if we get problems.

crates/pgt_workspace/src/features/completions.rs
juleswritescode: because I need to peek into the next statement for the overlap check :/

psteinroe: makes sense. maybe we can find a good abstraction for this but for now lets roll with this.

PR #313, opened 2025-04-04
fix: properly trim statement

juleswritescode:
Before, changing `select * from ` to `select * from` resultet in a `ModifiedStatement.new_stmt_text === 'select * fro` (the m was accidentally trimmed).

This was because we subtracted the diff_size from the end of the new statement, even if the diff range wasn't included in the new statement anymore.

This PR fixes it :)

### DIFF
--- a/crates/pgt_workspace/src/workspace/server/change.rs
+++ b/crates/pgt_workspace/src/workspace/server/change.rs
@@ -151,6 +151,8 @@ impl Document {
         let mut start = change_range.start();
         let mut end = change_range.end().min(content_size);
 
+        let is_trim = change_range.start() >= content_size;
+
         let mut affected_indices = Vec::new();
         let mut prev_index = None;
         let mut next_index = None;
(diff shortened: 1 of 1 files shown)


crates/pgt_workspace/src/workspace/server/change.rs
```diff
@@ -269,7 +276,9 @@ impl Document {
                 let new_range = new_ranges[0].add(affected_range.start());
                 let (old_id, old_range) = self.positions[affected_idx];
 
-                // move all statements after the afffected range
+                println!("{:?}", old_range);
+
```
juleswritescode, line 280:
```suggestion
```

crates/pgt_workspace/src/workspace/server/change.rs
```diff
@@ -1322,4 +1325,139 @@ mod tests {
...
+                assert_eq!(change_range, &TextRange::new(13.into(), 14.into()));
+                assert_eq!(change_text, "");
+                assert_eq!(new_stmt_text, "select * from");
+
+                // the whitespace was not considered
+                // to be a part of the statement
+                assert_eq!(old_stmt_text, "select * from");
+            }
+
+            _ => panic!("Did not yield a modified statement."),
```
psteinroe, line 1369: nit: I think asserting would be more idiomatic 
```suggestion
            _ => assert!(false, "Did not yield a modified statement."),
```

crates/pgt_workspace/src/workspace/server/change.rs
```diff
@@ -1322,4 +1325,139 @@ mod tests {
...
+                assert_eq!(change_range, &TextRange::new(12.into(), 14.into()));
+                assert_eq!(change_text, "");
+                assert_eq!(new_stmt_text, "select * fro");
+
+                // the whitespace was not considered
+                // to be a part of the statement
+                assert_eq!(old_stmt_text, "select * from");
+            }
+
+            _ => panic!("Did not yield a modified statement."),
```
psteinroe, line 1415:
```suggestion
            _ => assert!(false, "Did not yield a modified statement."),
```

crates/pgt_workspace/src/workspace/server/change.rs
```diff
@@ -1322,4 +1325,139 @@ mod tests {
...
+                    ..
+                } = stmt;
+
+                assert_eq!(change_range, &TextRange::new(9.into(), 11.into()));
+                assert_eq!(change_text, "");
+                assert_eq!(old_stmt_text, "select *   from users");
+                assert_eq!(new_stmt_text, "select * from users");
+            }
+
+            _ => panic!("Did not yield a modified statement."),
```
psteinroe, line 1458:
```suggestion
            _ => assert!(false, "Did not yield a modified statement."),
```

psteinroe (approved):
Great catch!

PR #315, opened 2025-04-04
doc: add zed installation instructions

jeffreyguenther:

It adds instructions for the Zed editor.

Docs don't exist.

## What is the new behavior?

N/A

## Additional context

N/A


psteinroe (approved):
well, that is straightforward. Thanks for doing it!

PR #322, opened 2025-04-09
refactor: parser

### LINKED ISSUE #178: support sql functions bodies
we should be able to extract the body of a `language sql` function as yet another statement and provide language intelligence features on it as usual.

psteinroe:
this rewires the workspace server for better DX and adds support for sql function bodies. 

closes #178

hear me out:

- the `Document` is responsible for extracting statements from a source file. right now, this is always SQL, but in the future we could make this generic. 
- the `Parser` is the new layer between the workspace server and the document. there is one parser per document, and the parser owns the document. 
- the parser is responsible for parsing the raw sql coming from the document, and is the main API layer for the workspace. 
- the consumer of the parser apis simply gets the statements and the related parse results. the consumer does not need to know anything about sub statements 
- I added a nice pattern where the consumer of the Parser can define a mapper and / or a filter and then simply calls `parser.iter(ExecuteStatementMapper, CursorPositionFilter::new(pos))` and gets exactly the data it needs. 
- I also optimised the parsing in general: the only thing that actually cares about modified statements is tree sitter. for pg_query, we always just remove and add when we receive a modify statement. hence, we can simply lazily parse all but tree sitter and just use modified / remove notifications to clear the cache. this also opened up the way for a nicer api because if you want the pg_query results, you always get Result<Node, SyntaxError> and we do not need to wrap it in an Option<> anymore.
- the statement id is now not just an usize anymore but an enum that can reflect sub statements. if we remove parse results for a parent, we also remove it for its sub statement.


crates/pgt_query_ext/src/lib.rs
```diff
@@ -30,3 +30,48 @@ pub fn parse(sql: &str) -> Result<NodeEnum> {
             .ok_or_else(|| Error::Parse("Unable to find root node".to_string()))
     })?
 }
+
+#[cfg(test)]
+mod tests {
```
psteinroe, line 35: clean this up

crates/pgt_workspace/src/workspace/server/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+        self.iter_with_filter(mapper, DefaultFilter)
+    }
+
+    pub fn iter_with_filter<'a, M, F>(&'a self, mapper: M, filter: F) -> ParseIterator<'a, M, F>
+    where
+        M: StatementMapper<'a>,
+        F: StatementFilter<'a>,
+    {
+        ParseIterator::new(self, mapper, filter)
+    }
```
psteinroe, line 120: this is the nice pattern

crates/pgt_workspace/src/workspace/server/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+        // Process the next top-level statement
+        let next_statement = self.statements.next();
+
+        if let Some((root_id, range, content)) = next_statement {
+            // If we should include sub-statements and this statement has an AST
+            let content_owned = content.to_string();
+            if let Ok(ast) = self
+                .parser
+                .ast_db
+                .load_parse(&root_id, &content_owned)
```
psteinroe, line 190: when statements are requested, we automatically handle sub statement and return them as any other statement

crates/pgt_workspace/src/workspace/server/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+        let ast_result = parser.ast_db.load_parse(&id, content);
+        let ast_option = match &*ast_result {
+            Ok(node) => Some(node.clone()),
+            Err(_) => None,
+        };
+
+        (id, range, content.to_string(), ast_option)
+    }
+}
+
```
psteinroe, line 262: every use case simply creates their own Mapper and defines exactly what data they need :)

crates/pgt_workspace/src/workspace/server/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub fn new(id: StatementId) -> Self {
+        Self { id }
+    }
+}
+
+impl<'a> StatementFilter<'a> for IdFilter {
+    fn apply(&self, id: &StatementId, _range: &TextRange) -> bool {
+        *id == self.id
+    }
+}
```
psteinroe, line 374: same goes for filters, where these two probably is all we need

crates/pgt_workspace/src/workspace/server.rs
```diff
@@ -307,20 +254,26 @@ impl Workspace for WorkspaceServer {
...
-        }
+        let actions = parser
+            .iter_with_filter(
+                DefaultMapper,
+                CursorPositionFilter::new(params.cursor_position),
+            )
+            .map(|(stmt, _, txt)| {
+                let title = format!(
+                    "Execute Statement: {}...",
+                    txt.chars().take(50).collect::<String>()
```
psteinroe, line 265: beautiful

crates/pgt_completions/src/complete.rs
```diff
@@ -14,7 +14,7 @@ pub struct CompletionParams<'a> {
     pub position: TextSize,
     pub schema: &'a pgt_schema_cache::SchemaCache,
     pub text: String,
-    pub tree: Option<&'a tree_sitter::Tree>,
```
psteinroe, line 17: tree sitter cannot fail and with the new pattern this doesnt need to be an option anymore

crates/pgt_workspace/src/workspace/server/parser.rs
psteinroe: edit: completions will need their own Filter :D

crates/pgt_workspace/src/workspace/server.rs
psteinroe: also works out of the box with sub statements like sql function bodies and the workspace implementation doesnt even know it

crates/pgt_workspace/src/workspace/server/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+use super::{
+    change::StatementChange,
+    document::{Document, StatementIterator},
+    pg_query::PgQueryStore,
+    sql_function::SQLFunctionBodyStore,
+    statement_identifier::StatementId,
+    tree_sitter::TreeSitterStore,
+};
+
+pub struct Parser {
```
juleswritescode, line 19: `Parser` is a bit of an odd name to me, since it isn't used to do parsing, but as more of a container for the results.

Is this maybe a `ParsedDocument` ? Or a `ParsedStatements` struct?

crates/pgt_workspace/src/workspace/server/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+        id: StatementId,
+        range: TextRange,
+        content: &str,
+    ) -> Self::Output {
+        let cst_result = parser.cst_db.load_parse(&id, content);
+        (id, range, content.to_string(), cst_result)
+    }
+}
+
+pub struct DefaultFilter;
```
juleswritescode, line 337: `NoopFilter` ?

crates/pgt_workspace/src/workspace/server/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+impl<'a> StatementFilter<'a> for DefaultFilter {
+    fn apply(&self, _id: &StatementId, _range: &TextRange) -> bool {
+        true
+    }
+}
+
+pub struct CursorPositionFilter {
+    pos: TextSize,
+}
```
juleswritescode, line 346: love this!
I'm already seeing a `CursorPositionWithOffsetFilter` for the completions

crates/pgt_workspace/src/workspace/server/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+        &self,
+        parser: &'a Parser,
+        id: StatementId,
+        range: TextRange,
+        content: &str,
+    ) -> Self::Output;
+}
+
+pub trait StatementFilter<'a> {
+    fn apply(&self, id: &StatementId, range: &TextRange) -> bool;
```
juleswritescode, line 141: Nit: If I were to implement a filter, I would have to dive into the source code of `ParseIterator` to find out whether this should return `true` or `false` when I want the statement included.
How about `filter_in` or `filter_out` or `predicate` ?

crates/pgt_workspace/src/workspace/server/tree_sitter.rs
```diff
@@ -20,24 +19,34 @@ impl TreeSitterStore {
...
         TreeSitterStore {
             db: DashMap::new(),
-            parser: RwLock::new(parser),
+            parser: Mutex::new(parser),
         }
     }
 
-    pub fn get_parse_tree(&self, statement: &Statement) -> Option<Arc<tree_sitter::Tree>> {
-        self.db.get(statement).map(|x| x.clone())
+    pub fn load_parse(&self, statement: &StatementId, content: &str) -> Arc<tree_sitter::Tree> {
```
juleswritescode, line 26: `get_or_cache_tree` ?

crates/pgt_workspace/src/workspace/server/tree_sitter.rs
```diff
@@ -20,24 +19,34 @@ impl TreeSitterStore {
 
         TreeSitterStore {
             db: DashMap::new(),
-            parser: RwLock::new(parser),
+            parser: Mutex::new(parser),
```
juleswritescode, line 22: nice that you changed this back.
I think in general we can opt for simplicity over performance/concurrency optimizations – there'll only ever be very few users

juleswritescode (commented):
Great PR! :) 
I was scared that accomodating nested statements would make things more complex, but this even makes it simpler!!

psteinroe:
@juleswritescode for me ParsedDocument or DocumentParse makes more sense than ParsedStatements

crates/pgt_workspace/src/workspace/server/parser.rs
juleswritescode: let's do that separately

juleswritescode (approved):

PR #336, opened 2025-04-14
fix: plpglsql cmds

### LINKED ISSUE #335: support psql commands
psql commands \if, \set etc are normal to see in .sql source files in order to support certain kinds of build managements systems.  Currently they are thrown out as unrecognized.  I guess this is a bit of a stretch to be called a bug, but in general the checker should parse .sql that psql can parse.  backslash commands are very simple; if the first character of a line is a backslash the rest of the line can be ignored.  I not believe there is any need to interpret the commands.

## To Reproduce

This file:
psql_test.sql

```
\if 1

do $$
begin
  raise notice '%', $q$
\this is in a string;
  $q$;

end;
$$;

    \endif
```

...yields...

```
./psql_test.sql:1:1 syntax ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Invalid statement: syntax error at or near "\"

  > 1 │ \if 1
      │ ^^^^^
    2 │
    3 │ do $$


./psql_test.sql:12:5 syntax ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Invalid statement: syntax error at or near "\"

    10 │ $$;·
    11 │
  > 12 │     \endif
       │     ^^^^^^
    13 │


Checked 1 file in 3ms. No fixes applied.
Found 2 errors.
check ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━


  ✖ Some errors were emitted while running checks.
```

psql however yields:

```
NOTICE:
\this is in a string;

DO
```

psteinroe:
fixes #335

### DIFF
--- a/crates/pgt_statement_splitter/src/parser/common.rs
+++ b/crates/pgt_statement_splitter/src/parser/common.rs
@@ -1,4 +1,4 @@
-use pgt_lexer::{SyntaxKind, Token, TokenType};
+use pgt_lexer::{SyntaxKind, Token, TokenType, WHITESPACE_TOKENS};
 
 use super::{
     Parser,
@@ -23,6 +23,13 @@ pub fn source(p: &mut Parser) {
                 ..
             } => {
                 p.advance();
+            }
+            Token {
+                kind: SyntaxKind::Ascii92,
+                ..
+            } => {
+                println!("current {:?}", p.current());
+                plpgsql_command(p);
             }
             _ => {
                 statement(p);
@@ -87,6 +94,24 @@ pub(crate) fn parenthesis(p: &mut Parser) {
     }
 }
 
+pub(crate) fn plpgsql_command(p: &mut Parser) {
+    p.expect(SyntaxKind::Ascii92);
+
+    loop {
+        match p.current().kind {
+            SyntaxKind::Newline => {
+                p.advance();
+                break;
+            }
+            _ => {
+                // advance the parser to the next token without ignoring irrelevant tokens
+                // we would skip a newline with `advance()`
+                p.current_pos += 1;
+            }
+        }
+    }
+}
+
 pub(crate) fn case(p: &mut Parser) {
     p.expect(SyntaxKind::Case);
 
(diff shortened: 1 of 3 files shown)


crates/pgt_statement_splitter/src/parser/common.rs
```diff
@@ -125,6 +149,36 @@ pub(crate) fn unknown(p: &mut Parser, exclude: &[SyntaxKind]) {
...
+                    .filter(|t| **t != SyntaxKind::Newline)
+                    .collect::<Vec<_>>();
+
+                // go back from the current position without ignoring irrelevant tokens
+                if p.tokens
+                    .iter()
+                    .take(p.current_pos)
+                    .rev()
+                    .find(|t| !irrelevant.contains(&&t.kind))
+                    .is_some_and(|t| t.kind == SyntaxKind::Newline)
```
juleswritescode, line 176: The `is_irrelevant` function in the parser already checks for `!= Newline`. Is that broken maybe?

juleswritescode: maybe `p.look_back()` just works

juleswritescode: Ah I see! thanks 💪🏻

crates/pgt_statement_splitter/src/parser/common.rs
```diff
@@ -125,6 +149,36 @@ pub(crate) fn unknown(p: &mut Parser, exclude: &[SyntaxKind]) {
...
+                ..
+            } => {
+                // pgsql commands e.g.
+                //
+                // ```
+                // \if test
+                // ```
+                //
+                // we wait for "\" and check if the previous token is a newline
+
```
juleswritescode, line 163:
```suggestion
                // we wait for "\" and check if the previous token is a single newline

```

juleswritescode (approved):

PR #337, opened 2025-04-14
fix: comments only

psteinroe:
fixes a special case where a sql file contains only sql comments and we start typing. it started to create statements out of nowhere because the full_affected_range is just the change_range. I fixed this by handling this special case separately.


juleswritescode (approved):
good catch 👍

PR #338, opened 2025-04-15
fix: release changelog

psteinroe:
- only generates the changelog for the version that is currently released 
- (I think) I fixed the deploy docs on release issue by allowing all tags to push to `github-pages `
Screenshot 2025-04-15 at 08 37 23
- makes the template a bit nicer. example from the latest release:

# Postgres Language Server

A collection of language tools and a Language Server Protocol (LSP) implementation for Postgres, focusing on developer experience and reliable SQL tooling.

### 🚀 Features

- Annotations (#331)
- Request autocompletion without typing a letter (#310)

### 🐛 Bug Fixes

- Make sure range is correct for eof deletions (#311)
- Properly trim statement (#313)
- Downgrade tunners (#318)
- Parse unions (#329)
- C-style comments (#332)

### 🚜 Refactor

- Parser (#322)
- Simplify parser ? (#330)

### 📚 Documentation

- Add zed installation instructions (#315)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -126,7 +126,7 @@ jobs:
         id: create_changelog
         with:
           config: cliff.toml
-          args: --bump
+          args: --bump --latest
         env:
           GITHUB_REPO: ${{ github.repository }}
 
--- a/cliff.toml
+++ b/cliff.toml
@@ -8,14 +8,14 @@
 [changelog]
 # template for the changelog header
 header = """
-# Changelog\n
-All notable changes to this project will be documented in this file.\n
+# Postgres Language Server\n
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -126,7 +126,7 @@ jobs:
         id: create_changelog
         with:
           config: cliff.toml
-          args: --bump
+          args: --bump --latest
```
juleswritescode, line 129: just FYI: I think this processes from the latest `release`, not `prerelease`. But that's what we want right?

juleswritescode (approved):

.github/workflows/release.yml
psteinroe: yes, I think that's what we want too. it worked fine locally at least.

PR #348, opened 2025-04-16
fix: report db connection errors

### LINKED ISSUE #344: Better reporting for database connection issues
# Bug report

test.sql:
INSERT INTO foo DEFAULT VALUES ON CONFLICT DO NOTHING;

...is parsing, but takes ~ 10 seconds to parse.  This is slow enough to be considered bug IMNSHO.

psteinroe:
this pr reports database errors back when running the type-checker. they are reported as `WorkspaceError` and abort the diagnostics pull. the idea is that the user directly sees what's wrong instead of just waiting for 10s to see only the syntax errors and then wondering why the performance is so bad.

fixes #344

### DIFF
--- a/crates/pgt_typecheck/src/lib.rs
+++ b/crates/pgt_typecheck/src/lib.rs
@@ -3,10 +3,9 @@ mod diagnostics;
 pub use diagnostics::TypecheckDiagnostic;
 use diagnostics::create_type_error;
 use pgt_text_size::TextRange;
-use sqlx::Executor;
-use sqlx::PgPool;
 use sqlx::postgres::PgDatabaseError;
 pub use sqlx::postgres::PgSeverity;
+use sqlx::{Executor, PgPool};
 
 #[derive(Debug)]
 pub struct TypecheckParams<'a> {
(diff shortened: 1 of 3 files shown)


crates/pgt_typecheck/src/lib.rs
```diff
@@ -39,28 +40,25 @@ pub async fn check_sql(params: TypecheckParams<'_>) -> Option<TypecheckDiagnosti
...
+    let mut conn = params.conn.acquire().await?;
 
     // Postgres caches prepared statements within the current DB session (connection).
     // This can cause issues if the underlying table schema changes while statements
     // are cached. By closing the connection after use, we ensure a fresh state for
     // each typecheck operation.
     conn.close_on_drop();
 
-    let res = conn.prepare(params.sql).await;
+    let res = params.conn.prepare(params.sql).await;
```
psteinroe, line 54:
```suggestion
    let res = conn.prepare(params.sql).await;
```

juleswritescode (approved):
very nice!!!

PR #358, opened 2025-04-21
fix(completions): remove , as trigger char

juleswritescode:
If `' '` triggers completions, `','` probably shouldn't.


psteinroe (approved):
Not sure about that but let's test it

PR #366, opened 2025-04-22
fix: sql fn params

### LINKED ISSUE #352: SQL function: False error reports for function parameter references
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application.
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com).

## Describe the bug

If a function takes in a named parameter and the parameter is used in the function body, the used parameter is flagged as a false error claiming that the column does not exist (But it is a function parameter).

Image

## To Reproduce

Create a function with a named parameter.

```
create or replace function users.select_no_ref (user_id uuid) returns table (
    first_name text
) language sql security invoker as $$
select
    first_name
FROM 
    users_hidden.users
where 
    id = user_id;
$$;
```

## Expected behavior

Named function parameters should be recognized as correct code.

## System information

- OS: windows
- code editor: vs-code
- pls version: 0.5.0

### LINKED ISSUE #353: SQL function: False error reports for function parameter references using explicit function name reference
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application.
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com).

## Describe the bug

If a function takes in a named parameter and the parameter is used in the function body and referenced explicitly using the function name, the used function name is flagged as a false error claiming an incorrect FROM clause (probably because it expects a Select * from <function_name> statement).

Image

## To Reproduce

Create a function with a named parameter and use a explicit function parameter reference.

```
create or replace function users.select_with_function_ref (user_id uuid) returns table (
    first_name text
) language sql security invoker as $$
select
    first_name
FROM 
    users_hidden.users
where 
    id = select_with_function_ref.user_id;
$$;
```

## Expected behavior

Named function parameters using an explicit function name as reference should be recognized as correct code.

## System information

- OS: windows
- code editor: vs-code
- pls version: 0.5.0

psteinroe:
the idea is to replace the fn params with a default value based on their type:
```sql
create or replace function users.select_no_ref (user_id uuid) returns table (
    first_name text
) language sql security invoker as $$
select
    first_name
FROM 
    users_hidden.users
where 
    id = user_id;
$$;
```

will become 

```sql
create or replace function users.select_no_ref (user_id int4) returns table (
    first_name text
) language sql security invoker as $$
select
    first_name
FROM 
    users_hidden.users
where 
    id = 0;
$$;
```

## Todo
- [ ] pass params to typechecker 
- [ ] implement  `apply_identifiers`

fixes #353 
fixes #352

### DIFF
--- a/crates/pgt_typecheck/src/typed_identifier.rs
+++ b/crates/pgt_typecheck/src/typed_identifier.rs
@@ -1,0 +1,18 @@
+#[derive(Debug)]
+pub struct TypedIdentifier {
+    pub schema: Option<String>,
+    pub relation: String,
+    pub name: String,
+    pub type_: String,
+}
+
+/// Applies the identifiers to the SQL string by replacing them with their default values.
+pub fn apply_identifiers<'a>(
+    identifiers: Vec<TypedIdentifier>,
+    schema_cache: &'a pgt_schema_cache::SchemaCache,
+    cst: &'a tree_sitter::Node<'a>,
+    sql: &'a str,
+) -> &'a str {
+    // TODO
+    sql
+}
--- a/crates/pgt_workspace/src/workspace/server/sql_function.rs
+++ b/crates/pgt_workspace/src/workspace/server/sql_function.rs
@@ -6,6 +6,18 @@ use pgt_text_size::TextRange;
 use super::statement_identifier::StatementId;
 
 #[derive(Debug, Clone)]
+pub struct SQLFunctionArgs {
+    pub name: Option<String>,
+    pub type_: (Option<String>, String),
+}
+
+#[derive(Debug, Clone)]
+pub struct SQLFunctionSignature {
+    pub name: (Option<String>, String),
+    pub args: Vec<SQLFunctionArgs>,
+}
+
+#[derive(Debug, Clone)]
 pub struct SQLFunctionBody {
     pub range: TextRange,
     pub body: String,
@@ -13,11 +25,33 @@ pub struct SQLFunctionBody {
 
 pub struct SQLFunctionBodyStore {
     db: DashMap<StatementId, Option<Arc<SQLFunctionBody>>>,
+    sig_db: DashMap<StatementId, Option<Arc<SQLFunctionSignature>>>,
 }
 
 impl SQLFunctionBodyStore {
     pub fn new() -> SQLFunctionBodyStore {
-        SQLFunctionBodyStore { db: DashMap::new() }
+        SQLFunctionBodyStore {
+            db: DashMap::new(),
+            sig_db: DashMap::new(),
+        }
+    }
+
+    pub fn get_function_signature(
+        &self,
+        statement: &StatementId,
+        ast: &pgt_query_ext::NodeEnum,
+    ) -> Option<Arc<SQLFunctionSignature>> {
+        // First check if we already have this statement cached
+        if let Some(existing) = self.sig_db.get(statement).map(|x| x.clone()) {
+            return existing;
+        }
+
+        // If not cached, try to extract it from the AST
+        let fn_sig = get_sql_fn_signature(ast).map(Arc::new);
+
+        // Cache the result and return it
+        self.sig_db.insert(statement.clone(), fn_sig.clone());
+        fn_sig
     }
 
     pub fn get_function_body(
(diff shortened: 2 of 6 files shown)


crates/pgt_workspace/src/workspace/server.rs
```diff
@@ -365,19 +370,40 @@ impl Workspace for WorkspaceServer {
...
         {
             let path_clone = params.path.clone();
+            let schema_cache = self.schema_cache.load(pool.clone())?;
+            let schema_cache_arc = Arc::new(schema_cache.as_ref().clone());
             let input = parser.iter(AsyncDiagnosticsMapper).collect::<Vec<_>>();
+            // sorry for the ugly code :(
             let async_results = run_async(async move {
                 stream::iter(input)
-                    .map(|(_id, range, content, ast, cst)| {
+                    .map(|(_id, range, content, ast, cst, sign)| {
```
juleswritescode, line 379: should we rename to `fn_signature`?

crates/pgt_workspace/src/workspace/server/sql_function.rs
```diff
@@ -1,56 +1,80 @@
...
+pub struct SQLFunctionArgs {
+    pub name: Option<String>,
+    pub type_: ArgType,
 }
 
-pub struct SQLFunctionBodyStore {
-    db: DashMap<StatementId, Option<Arc<SQLFunctionBody>>>,
+#[derive(Debug, Clone)]
+pub struct SQLFunctionSignature {
+    pub name: (Option<String>, String),
```
juleswritescode, line 18: nit: i think it would be more readable if it had a `name` and a `schema` property

crates/pgt_workspace/src/workspace/server/sql_function.rs
```diff
@@ -1,56 +1,80 @@
...
+    pub schema: Option<String>,
+    pub name: String,
+    pub is_array: bool,
+}
 
 #[derive(Debug, Clone)]
-pub struct SQLFunctionBody {
-    pub range: TextRange,
-    pub body: String,
+pub struct SQLFunctionArgs {
```
juleswritescode, line 11: this is a single `Arg` right?

crates/pgt_workspace/src/workspace/server/sql_function.rs
```diff
@@ -1,56 +1,80 @@
...
+pub fn get_sql_fn_signature(ast: &pgt_query_ext::NodeEnum) -> Option<SQLFunctionSignature> {
+    let create_fn = match ast {
+        pgt_query_ext::NodeEnum::CreateFunctionStmt(cf) => cf,
+        _ => return None,
+    };
 
-        // If not cached, try to extract it from the AST
-        let fn_body = get_sql_fn(ast, content).map(Arc::new);
+    // Extract language from function options
+    let language = find_option_value(create_fn, "language")?;
```
juleswritescode, line 36: i really like that api

crates/pgt_typecheck/src/typed_identifier.rs
```diff
@@ -0,0 +1,340 @@
...
+}
+
+/// Format the default value based on the type and whether it's an array
+fn get_formatted_default_value(pg_type: &PostgresType, is_array: bool) -> String {
+    // Get the base default value for this type
+    let default = resolve_default_value(pg_type);
+
+    let default = if default.len() > "NULL".len() {
+        // If the default value is longer than "NULL", use "NULL" instead
+        "NULL".to_string()
```
juleswritescode, line 86: why are we doing this?

crates/pgt_typecheck/src/typed_identifier.rs
```diff
@@ -0,0 +1,340 @@
...
+            // Resolve the type based on whether we're accessing a field of a composite type
+            let type_ = resolve_type(identifier, position, &parts, schema_cache)?;
+
+            Some((m.get_byte_range(), type_, identifier.type_.is_array))
+        })
+        .collect();
+
+    let mut result = sql.to_string();
+
+    let mut valid_positions = true;
```
juleswritescode, line 55: I'm not sure what `position` is in this context.
I understand it this way: We'll need to adjust the statements ranges if a parameter's default value is longer than the substituted text.
If that's correct, how about we rename this to something like `should_adjust_ranges` (default false) ?

crates/pgt_typecheck/src/typed_identifier.rs
```diff
@@ -0,0 +1,340 @@
...
+fn get_formatted_default_value(pg_type: &PostgresType, is_array: bool) -> String {
+    // Get the base default value for this type
+    let default = resolve_default_value(pg_type);
+
+    let default = if default.len() > "NULL".len() {
+        // If the default value is longer than "NULL", use "NULL" instead
+        "NULL".to_string()
+    } else {
+        // Otherwise, use the default value
+        default
```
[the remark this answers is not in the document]
psteinroe: 🫡

crates/pgt_typecheck/src/typed_identifier.rs
```diff
@@ -0,0 +1,340 @@
...
+    // Case 2: Named reference (e.g., fn_name.custom_type.v_test2)
+    identifiers.iter().find_map(|identifier| {
+        let name = identifier.name.as_ref()?;
+
+        parts
+            .iter()
+            .enumerate()
+            .find(|(_idx, part)| **part == name)
+            .map(|(idx, _)| (identifier, idx))
+    })
```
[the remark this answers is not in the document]
psteinroe: I think `schema.fn_name.type_name.field_name` would be possible but I have never seen it ever in the wild

crates/pgt_typecheck/src/typed_identifier.rs
psteinroe: sometimes (almost never), we get a position back for the type error. in that case, we show the diagnostics on that position instead of the entire statement. since we are replacing the vars in the text, the position we get back does not match the position in the original statement anymore *if the replacements are not the same length*. hence, if `valid_positions` is `false`, we throw away the position from the type error and always show the error on the entire length of the statement.

crates/pgt_typecheck/src/typed_identifier.rs
psteinroe: because we need to throw away positions when the replacement is longer than its original, and `NULL` will be shorter than a few default values (e.g. timestamps or custom enums)

crates/pgt_typecheck/src/typed_identifier.rs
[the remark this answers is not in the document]
juleswritescode: ack then let's ignore it for now!  plus, it seems to be an easy fix.

PR #370, opened 2025-04-23
feat(completions): insert schema name when selecting non-public tables

juleswritescode:
yayy

https://github.com/user-attachments/assets/4711f5a6-fa9b-425e-9e75-7b4a06e4f6a0


psteinroe (approved):
Nice feature!!

PR #373, opened 2025-04-25
fix: timestamp with time zone split

psteinroe:


juleswritescode (approved):
Nice! 

Linting issues look like the same I had earlier, should be fixed after [this](https://github.com/supabase-community/postgres-language-server/pull/374) is merged ❤️

PR #374, opened 2025-04-25
chore: align ci with local dev

juleswritescode:
- pin the rust toolchain to the version we use for the workspace
- reuse `just` commands in CI
- add steps where we print the versions of the tools used in CI
- install `sqlx-cli` with `just install-tools` and `just upgrade-tools` 
- remove `bun install` from `just upgrade-tools` 
- fixed all issues from clippy

the `rust-toolchain.toml` file was weird – it complained for the longest time that `cargo fmt` wasn't available with the 1.86.0 version, but now it just works. Seems to be just buggy sometimes

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -33,8 +33,10 @@ jobs:
     steps:
       - name: Checkout PR branch
         uses: actions/checkout@v4
+
       - name: Free Disk Space
         uses: ./.github/actions/free-disk-space
+
       - name: Install toolchain
         uses: moonrepo/setup-rust@v1
         with:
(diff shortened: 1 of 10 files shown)


psteinroe (approved):

willruggiano:
Have you guys ever dabbled with Nix? I have found that it excels at this sort of thing (i.e. dev/ci determinism/reproducibility).

There is a [great blog post](https://mitchellh.com/writing/nix-with-dockerfiles) that was written primarily about building Docker images using Nix, but the sentiment applies to CI in general. But alas I am heavily invested and thus heavily biased :D

Anyways, if you are interested I'd be more than happy to discuss in detail with you. Or even if you aren't but are nonetheless curious! Thought I'd toss it out there as you are neck deep in a CI refactor.

PR #375, opened 2025-04-25
fix: do not attempt db connection if jsonc section is missing

juleswritescode:
If the user did not specify the `db` section in the jsonc, we would per default try to connect to `postgresql://postgres:postgres@127.0.0.1:5432/postgres`. 

The reason is the `biome_serialize_macros::Merge` macro:

If two `configuration.db`s are `None`, they are merged into a `Some(PartialConfiguration {})` where all properties are `None`; if we then call `DatabaseSettings::from(partial_configuration)`, the partial config is merged with `DatabaseSettings::default()` – where we specified the above connection settings.

We can't really work around that without refactoring the whole configuration shenaigans, so I pulled the `cli_options.skip_db` flag into the database settings. 

You can still pass it as `postgrestools check --disable-db test.sql`, and it is hidden in the `.jsonc`. 

When a user does not specify a `db.host` in the `jsonc`, all db-connections are now disabled.

### DIFF
--- a/packages/@postgrestools/backend-jsonrpc/src/workspace.ts
+++ b/packages/@postgrestools/backend-jsonrpc/src/workspace.ts
@@ -179,21 +179,36 @@ export interface GetCompletionsParams {
 	 */
 	position: TextSize;
 }
-export interface CompletionResult {
+export interface CompletionsResult {
 	items: CompletionItem[];
 }
 export interface CompletionItem {
+	completion_text?: CompletionText;
 	description: string;
 	kind: CompletionItemKind;
 	label: string;
 	preselected: boolean;
-	score: number;
+	/**
+	 * String used for sorting by LSP clients.
+	 */
+	sort_text: string;
 }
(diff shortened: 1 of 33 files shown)


packages/@postgrestools/backend-jsonrpc/src/workspace.ts
```diff
@@ -179,21 +179,36 @@ export interface GetCompletionsParams {
 	 */
 	position: TextSize;
```
juleswritescode, line 180: auto-generated

postgrestools.jsonc
```diff
@@ -17,7 +17,7 @@
   // YOU CAN COMMENT ME OUT :)
   "db": {
     "host": "127.0.0.1",
-    "port": 5432,
+    "port": 54322,
```
juleswritescode, line 20:
```suggestion
    "port": 5432,
```

psteinroe (approved):

PR #380, opened 2025-04-30
fix?(splitter): allow newlines after commas

juleswritescode:
looks like we can assume that it's always a statement if there's a comma before the newlines? 

```sql
select
            email,



        from
            auth.users;
```

Note that this doesn't fix [this one](https://github.com/supabase-community/postgres-language-server/issues/372) – it's apparently not a splitter-problem, the file is correclty parsed as a single statement

### DIFF
--- a/crates/pgt_completions/src/builder.rs
+++ b/crates/pgt_completions/src/builder.rs
@@ -86,7 +86,7 @@ fn should_preselect_first_item(items: &Vec<PossibleCompletionItem>) -> bool {
     let second = items_iter.next();
 
     first.is_some_and(|f| match second {
-        Some(s) => (f.score.get_score() - s.score.get_score()) > 10,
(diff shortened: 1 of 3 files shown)


crates/pgt_completions/src/builder.rs
```diff
@@ -86,7 +86,7 @@ fn should_preselect_first_item(items: &Vec<PossibleCompletionItem>) -> bool {
     let second = items_iter.next();
 
     first.is_some_and(|f| match second {
-        Some(s) => (f.score.get_score() - s.score.get_score()) > 10,
+        Some(s) => (f.score.get_score() - s.score.get_score()) > 15,
         None => true,
-    })
+    }) && items.len() >= 10
```
juleswritescode, line 91: just some minor tweaking here

psteinroe (approved):
great idea, thanks!!

PR #389, opened 2025-05-05
fix: properly handle multiple changes at once

psteinroe:
multiple changes at once happen during search and replace operations. 

Right now, we are handling each individually, one after another. this breaks starting at the second change because the range of each change is always related to the original state, not the previous. after fiddling with the `apply_change` method I figured this solution is the easiest. it expects that all changes are ordered properly, but afaik that's reasonable. if not and we receive bug reports, we can simply sort them. 

EDIT: ahh I need to handle the deletions case too! ✅

### DIFF
--- a/crates/pgt_workspace/src/workspace/server/change.rs
+++ b/crates/pgt_workspace/src/workspace/server/change.rs
@@ -1,9 +1,5 @@
-use biome_deserialize::Text;
 use pgt_text_size::{TextLen, TextRange, TextSize};
-use std::{
-    i64,
-    ops::{Add, Sub},
-};
+use std::ops::{Add, Sub};
 
 use crate::workspace::{ChangeFileParams, ChangeParams};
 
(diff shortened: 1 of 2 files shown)


crates/pgt_workspace/src/workspace/server/change.rs
```diff
@@ -63,11 +63,33 @@ impl Document {
...
+        for change in &change.changes {
+            let adjusted_change = if offset != 0 && change.range.is_some() {
+                &ChangeParams {
+                    text: change.text.clone(),
+                    range: change.range.map(|range| {
+                        let start = u32::from(range.start());
+                        let end = u32::from(range.end());
+                        TextRange::new(
+                            TextSize::from((start as i64 + offset).try_into().unwrap_or(0)),
+                            TextSize::from((end as i64 + offset).try_into().unwrap_or(0)),
```
juleswritescode, line 81: nit: isn't this `range.checked_add(offset)`?

juleswritescode (approved):
Good catch, jesus! :D

crates/pgt_workspace/src/workspace/server/change.rs
psteinroe: The problem with checked_add is that it expects a TextSize which is always > 0.

PR #390, opened 2025-05-06
fix(completions): complete right columns right after JOIN ON

juleswritescode:
Before, we were only completing columns in a JOIN in a "Binary Expression".
But for TreeSitter to recognize a Binary Expression, we need an `=` sign. 

So this wouldn't complete, since there's no `=` sign:
`select * from users u join posts p on u.{}` 

Now, we're checking for the `ON` node, and completing columns if we're after it.

Note: Multiple joins are each individual join clauses with their own separate ON nodes, so there's no intereference:

```sql
select * 
from 
users u
join posts p on u.id = post.user_id -- join clause 1
join {} -- join clause 2
```

We won't autocompete a column, since there's no ON node in join cluase 2 🥳

### DIFF
--- a/crates/pgt_completions/src/context.rs
+++ b/crates/pgt_completions/src/context.rs
@@ -341,7 +341,7 @@ impl<'a> CompletionContext<'a> {
         node: tree_sitter::Node<'a>,
         cursor: &mut tree_sitter::TreeCursor<'a>,
     ) -> Option<WrappingClause<'a>> {
-        return match node.kind() {
+        match node.kind() {
             "where" => Some(WrappingClause::Where),
             "update" => Some(WrappingClause::Update),
             "select" => Some(WrappingClause::Select),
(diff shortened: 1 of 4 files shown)


crates/pgt_completions/src/context.rs
```diff
@@ -353,12 +308,67 @@ impl<'a> CompletionContext<'a> {
...
+    ) -> Option<WrappingClause<'a>> {
+        return match node.kind() {
+            "where" => Some(WrappingClause::Where),
+            "update" => Some(WrappingClause::Update),
+            "select" => Some(WrappingClause::Select),
+            "delete" => Some(WrappingClause::Delete),
+            "from" => Some(WrappingClause::From),
+            "join" => {
+                // sadly, we need to manually iterate over the children –
+                // `node.child_by_field_id(..)` does not work as expected
```
juleswritescode, line 352: we might wanna fix this upstream at some point

psteinroe (approved):

PR #397, opened 2025-05-12
feat(completions): complete policies

juleswritescode:
Enables completions for policies – it requires a bit more complexity because tree sitter can't parse SQL statements.

Sadly, completing `alter policy "{}" on public.users;` doesn't work, because an `""` will cause the scanner to fail.

https://github.com/user-attachments/assets/54bad4f0-9f4d-4941-b5ef-fa3a16c1f9a6


psteinroe (approved):
we can merge this, although I think the effort / complexity is not worth it :D

PR #400, opened 2025-05-20
feats(completions): complete insert, drop/alter table, ignore many situations, improve WHERE

juleswritescode:
I got carried away a bit!

This PR adds support for INSERT INTO completions and for DROP TABLE/ALTER TABLE statements.

It also flips the completions filtering from disallowlisting to allowlisting, so we'll have way fewer unwanted completions. 

https://github.com/user-attachments/assets/b5ddafc5-3525-46bc-a6d7-bec3b4a7c068

### DIFF
--- a/crates/pgt_completions/src/sanitization.rs
+++ b/crates/pgt_completions/src/sanitization.rs
@@ -193,11 +193,6 @@ fn cursor_before_semicolon(tree: &tree_sitter::Tree, position: TextSize) -> bool
         return false;
     }
 
-    // not okay to be on the semi.
-    if byte == leaf_node.start_byte() {
-        return false;
-    }
-
     leaf_node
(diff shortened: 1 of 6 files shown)


crates/pgt_completions/src/sanitization.rs
```diff
@@ -317,4 +324,18 @@ mod tests {
         assert!(cursor_before_semicolon(&tree, TextSize::new(16)));
         assert!(cursor_before_semicolon(&tree, TextSize::new(17)));
     }
+
+    #[test]
+    fn between_parentheses() {
+        let input = "insert into instruments ()";
+
+        // insert into (|) <- right in the parentheses
```
juleswritescode, line 332: that must hurt

PR #408, opened 2025-05-26
feat: workspace support

### LINKED ISSUE #334: Workspace support
tbd

psteinroe:
adds workspace support as well as support for `extends` in the config. This now works:

`./postgrestools.jsonc`
```jsonc
{
    // ...my linter config
}
```

`./project_one/postgrestools.jsonc`
```jsonc
{
    "extends": "../postgrestools.jsonc"
    // e.g. add specific db config for this project
}
```

Will need to check if client-side changes are required to enable workspace support there.

A few internal changes:
- we now store both the schema cache and the db connection per connection and not globally
- the connections have a idle timeout of a few minutes after which we drop the connection from the cache
- we select the current project when the user opens a file. this is the same pattern biome uses and I think it makes sense.
- EDIT: I want to check if we even need to track the current project since all requests should include the path from which we can find the correct one -> we cannot easily do that. I would go with the current approach for now and maybe do it in a follow up with a better analysis. 
- EDIT: Will fix windows path issues for tests 🥲 -> done 
- EDIT: Will remove all the windows debug logs 🥲 -> done 

ToDo:
- [x] add methods to workspace api
- [x] migrate schema cache manager to store per-connection
- [x] integration ls
- [x] integration cli
- [x] make test green
- [x] implement `try_from_payload` with extends
- [x] add test for extends config

closes #334


crates/pgt_workspace/src/workspace/server.rs
```diff
@@ -155,31 +199,62 @@ impl WorkspaceServer {
...
+        debug!(
+            "Compare the current project with the new one {:?} {:?} {:?}",
+            current_project_path,
+            params.path.as_ref(),
+            current_project_path.as_deref() != params.path.as_ref()
+        );
+
+        let is_new_path = match (current_project_path.as_deref(), params.path.as_ref()) {
+            (Some(current_project_path), Some(params_path)) => current_project_path != params_path,
+            _ => true,
```
juleswritescode, line 216: can we have a current_project_path but params_path? Then it wouldn't be a new path, right?

crates/pgt_workspace/src/settings.rs
```diff
@@ -17,9 +18,187 @@ use pgt_configuration::{
...
+
+impl WorkspaceSettings {
+    pub fn get_current_project_key(&self) -> ProjectKey {
+        self.current_project
+    }
+
+    pub fn get_current_project_path(&self) -> Option<&PgTPath> {
+        trace!("Current key {:?}", self.current_project);
+        let data = self.data.get(self.current_project);
+        if let Some(data) = data {
```
juleswritescode, line 55: isn't this `data.as_ref().map(|d| d.path)` ?

crates/pgt_fs/src/fs.rs
```diff
@@ -88,6 +89,11 @@ pub trait FileSystem: Send + Sync + RefUnwindSafe {
             // Iterate all possible file names
             for file_name in file_names {
                 let file_path = curret_search_dir.join(file_name);
+                println!(
```
juleswritescode, line 92: sanity check: might this crash the LSP because it prints to stdout?

crates/pgt_workspace/src/workspace/server/schema_cache_manager.rs
```diff
@@ -1,97 +1,36 @@
...
-                tracing::info!("Refreshed connection.");
-            }
+        if let Some(cache) = self.schemas.get(&key) {
+            return Ok(Arc::clone(&*cache));
         }
 
-        Ok(SchemaCacheHandle::new(&self.inner))
-    }
-}
+        let schema_cache = Arc::new(run_async(async move { SchemaCache::load(&pool).await })??);
```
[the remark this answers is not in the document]
psteinroe: good catch!

crates/pgt_workspace/src/workspace/server.rs
psteinroe: yeah, added it. although I am not sure when this an happen :D

crates/pgt_fs/src/fs/memory.rs
```diff
@@ -227,6 +228,14 @@ impl FileSystem for MemoryFileSystem {
 
         Ok(cb())
     }
+
+    fn resolve_configuration(
+        &self,
+        _specifier: &str,
+        _path: &Path,
+    ) -> Result<Resolution, ResolveError> {
+        todo!()
```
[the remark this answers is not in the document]
psteinroe: oxc_resolver doesnt work on the memory file system afaik and we dont need it. added a comment

crates/pgt_fs/src/fs.rs
psteinroe: good catch!

crates/pgt_workspace/src/workspace/server.rs
```diff
@@ -334,16 +409,19 @@ impl Workspace for WorkspaceServer {
         &self,
         params: PullDiagnosticsParams,
     ) -> Result<PullDiagnosticsResult, WorkspaceError> {
-        let settings = self.settings();
+        let settings = self.workspaces();
+        // TODO: not sure how we should handle this
+        // maybe fallback to default settings? or return an error?
+        let settings = settings.settings().expect("Settings should be initialized");
```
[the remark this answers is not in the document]
psteinroe: yeah, makes sense. added that.

crates/pgt_fs/src/fs/memory.rs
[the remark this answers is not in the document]
juleswritescode: I think `unreachable!()` is cool for these cases :)

PR #424, opened 2025-06-09
feat(cache): include procedures & aggregates, fix query

juleswritescode:
This adds procedures and aggregates to the schema cache.

It also fixes a bug where function arguments where sometimes in the wrong order due to using `unnest` (apparently that's a common issue?).


psteinroe (approved):

crates/pgt_schema_cache/src/functions.rs
```diff
@@ -115,3 +144,115 @@ impl SchemaCacheItem for Function {
...
+        assert_eq!(foo_fn.behavior, Behavior::Volatile);
+
+        // Find and check the procedure
+        let proc_fn = cache
+            .functions
+            .iter()
+            .find(|f| f.name == "my_cool_proc")
+            .unwrap();
+
+        println!("{:?}", proc_fn);
```
juleswritescode:
```suggestion
```

PR #428, opened 2025-06-15
fix: batched changes

psteinroe:
this should *finally* fix batched changes.

i was super confused every time i worked on this because it seemed like whatever i did, somehow something did not work. Turns out, we cannot rely on lsp clients to send batched changes consistently. this now fixes it by
1. sorting changes starting with the last
2. applying them starting from the last
3. not changing the ranges in any way

had to delete another test that has faulty data, but i am very sure that the three test cases cover both neovim and vscode behavior.

will push out a patch release once this is merged and hope the reports are not popping up again.


juleswritescode (approved):
nice! seems simpler, too :)

PR #430, opened 2025-06-19
feat(linting): customize severity per rule

juleswritescode:
That was fun, messing with the codegen and macros :)

Introduces a new required `severity` property on lint rules. It's still overwriteable via the `.jsonc` file. 

Also added two new squawk rules to make sure that the process works as expected.


xtask/codegen/src/lib.rs
```diff
@@ -84,5 +85,9 @@ pub enum TaskCommand {
         /// Group of the rule
         #[bpaf(long("group"))]
         group: String,
+
+        /// Group of the rule
```
juleswritescode:
```suggestion
        /// Severity of the rule
```

PR #447, opened 2025-07-11
refactor: drop change.rs <3

psteinroe:
patrick-zen png

as per PR title. notable changes:
- drop `change.rs` in favour of a simple `update_document` api that replaces the content
- simplify `StatementId` to just be a simple wrapper around the actual statement string. The content is wrapped in an `Arc<str>` to make cloning cheap. turns out string interning only makes sense when we have a lot of duplicate strings. `Arc<str>` is much more efficient and simpler in our case.
-  Remove `document.rs` and merge it into `parsed_document` to then rename it to `Document`
- replaces the `strum` usage to get the available command because it didn't feel right to make an empty string the default for a StatementId. replaced it with a simple manual implementation.

I ran a few benchmarks and the statement splitter performance seems to be good enough to run it on every keystroke on the entire file. 
```sh
large statement with length 863000                                                                            
                        time:   [23.180 ms 23.278 ms 23.386 ms]

small statement with length 19000                                                                            
                        time:   [886.69 µs 890.91 µs 896.35 µs]
```

ToDo
- [ ] add some tests
- [x] cleanup


crates/pgt_workspace/src/workspace/server/statement_identifier.rs
```diff
@@ -35,91 +17,83 @@ impl From<usize> for RootId {
...
-    }
-
-    pub fn is_child(&self) -> bool {
-        matches!(self, StatementId::Child(_))
+    /// Creates a child statement ID with the given content and parent content.
+    pub fn new_child(child_content: &str, parent_content: &str) -> Self {
+        StatementId::Child {
+            content: child_content.into(),
+            parent_content: parent_content.into(),
+        }
```
juleswritescode: Is this used/needed? feels like we could create orphaned childs with this – not a problem right now, but maybe leads to confusion in the future?

crates/pgt_workspace/src/workspace/server/statement_identifier.rs
```diff
@@ -35,91 +17,83 @@ impl From<usize> for RootId {
...
+            StatementId::Root { content } => StatementId::Child {
+                content: child_content.into(),
+                parent_content: content.clone(),
             },
+            StatementId::Child { .. } => panic!("Cannot create child from a child statement id"),
         }
     }
 
-    pub fn parent(&self) -> Option<StatementId> {
+    pub fn content(&self) -> &str {
```
juleswritescode, line 57: love it

crates/pgt_workspace/src/workspace/server.rs
```diff
@@ -641,9 +637,9 @@ impl Workspace for WorkspaceServer {
                 });
 
                 tracing::debug!(
-                    "Found {} completion items for statement with id {}",
+                    "Found {} completion items for statement with id {:?}",
```
juleswritescode: that's probably a lot of log output – not sure we even need the print

crates/pgt_lsp/src/capabilities.rs
```diff
@@ -67,3 +61,13 @@ pub(crate) fn server_capabilities(capabilities: &ClientCapabilities) -> ServerCa
         ..Default::default()
     }
 }
+
+/// Returns all available command IDs for capability registration.
+/// Since CommandActionCategory has variants with data, we can't use strum::IntoEnumIterator.
+/// Instead, we manually list the command IDs we want to register.
+fn available_command_ids() -> Vec<String> {
+    vec![
+        "postgres-tools.executeStatement".to_string(),
```
juleswritescode: The ID in `pgt_lsp/src/handlers/code_actions.rs` also needs to be adjusted 👍

juleswritescode: Also, we're now just sending the full statement over to the client anyway, so we might as well change the CommandActionCategory::ExecuteStatement(_) to hold a `String` instead of a `StatementId` – I think you can then use strum again :)

psteinroe: ah good point!

crates/pgt_workspace/src/workspace/server/statement_identifier.rs
psteinroe: nope, thanks!

crates/pgt_lsp/src/capabilities.rs
psteinroe: reverted my changes. I think its simpler to just add a comment to the default implementation

PR #448, opened 2025-07-12
refactor: drop change.rs <3

psteinroe:


juleswritescode (approved):
Nice :) 

> Document-level caches would accumulate indefinitely as users edited files.

psteinroe:
> Nice :)
> 
> > Document-level caches would accumulate indefinitely as users edited files.
> 
> I'm curious, is this a bug with the lib? Or what happens there?

this is just a follow-up of the change refactoring. before, we were removing data from the parse caches (`pg_query`, `treesitter`) based on the statement changed and statement deleted events that the change handling emitted. Without that, we never delete any parsing results. if a user now starts typing, we keep every parse result of every statement change. this pr fixes that with the LRU cache. 

I then researched a bit about dash map and replaced it in other places with a more suitable store.

PR #449, opened 2025-07-13
chore: expose schema_cache & file_context in lint rules

juleswritescode:
- The `Analyser::run` now takes `Vec<AnalysableStatement>` and an `Option<&SchemaCache>`
- it also initializes an `AnalysedFileContext`
- the `SchemaCache` and the `AnalysedFileContext` are passed into each linting rule.

This should roughly match how the [eugene](https://github.com/kaaveland/eugene) code base does it.

The rest of the changes are adaptions to those changes or minor things, e.g. a new document-mapper or some renamings.

- [x] there's a bug with the lint spans, I'll fix it and write a test for it


crates/pgt_analyser/src/lint/safety/adding_required_field.rs
```diff
@@ -27,7 +30,11 @@ declare_lint_rule! {
 impl Rule for AddingRequiredField {
     type Options = ();
 
-    fn run(ctx: &RuleContext<Self>) -> Vec<RuleDiagnostic> {
+    fn run(
+        ctx: &RuleContext<Self>,
+        _file_context: &AnalysedFileContext,
+        _schema_cache: Option<&SchemaCache>,
```
psteinroe: from an API standpoint, I would put these two into the `RuleContext`. After all, they are "context" info.

crates/pgt_analyse/src/analysed_file_context.rs
```diff
@@ -0,0 +1,7 @@
+#[derive(Default)]
+pub struct AnalysedFileContext {}
```
psteinroe, line 2: what is this for?

psteinroe (commented):
any reason why we do not simply put these into the rule context? from an API standpoint, I feel like they should?

crates/pgt_analyse/src/analysed_file_context.rs
juleswritescode: That's just a scaffold for the file context. 
We can add properties in future linting PRs. 
eugene works the same way: [link](https://github.com/kaaveland/eugene/blob/7f8b343ae93ed72692cf63ce9a4490fd617a7118/eugene/src/lints.rs#L120)

crates/pgt_analyser/src/lint/safety/adding_required_field.rs
juleswritescode: I felt the `RuleContext` held static metadata about the rule, while the other stuff is related to the db/data model and the parsed sql file.

But I don't have a hard opinion

PR #462, opened 2025-07-18
fix: `files.ignore` setting should work

juleswritescode:
So I took a look at the problem at hand. 

Here it is:

image

I then went back to my garage and sifted through my tools – until I found it.

…
…
…

The perfect tool for the job.

…
…
…

My choice came down to this:

image

…
…
…

It does the job just right.

…
…
…

Okay, jokes aside, I just wanted to write a macro as an exercise – we can also just add a couple lines to every workspace function 😉

### DIFF
--- a/crates/pgt_lsp/src/handlers/text_document.rs
+++ b/crates/pgt_lsp/src/handlers/text_document.rs
@@ -1,12 +1,10 @@
-use crate::{
-    diagnostics::LspError, documents::Document, session::Session, utils::apply_document_changes,
-};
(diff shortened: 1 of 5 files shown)


psteinroe (approved):
as the guy that spent month trying to get `change.rs` to work I have no right to not approve this monster.

juleswritescode:
> as the guy that spent month trying to get change.rs to work I have no right to not approve this monster.

when the macro becomes annoying to maintain, let's not feel bad and get rid of it quickly 😛

PR #464, opened 2025-07-18
fix: splitter crash

psteinroe:
fixes a crash in the splitter when the last char is a closing parentheses.


crates/pgt_statement_splitter/src/splitter/common.rs
```diff
@@ -70,7 +70,9 @@ pub(crate) fn parenthesis(p: &mut Splitter) {
                 depth += 1;
             }
             SyntaxKind::R_PAREN | SyntaxKind::EOF => {
-                p.advance();
+                if p.current() == SyntaxKind::R_PAREN {
+                    p.advance();
+                }
```
psteinroe, line 75: this is the actual fix :D

PR #466, opened 2025-07-19
refactor: extract completions context into `pgt_treesitter` crate

juleswritescode:
I wanted to extract the treesitter-context that's used in completions, so it can also be used for the on-hover feature.

I've also tried to clean up the APIs, detangled completion-context and sanitization, and extracted the general test helpers into the `pgt_test_utils` crate.


psteinroe (approved):
nice!!

Issue #454, opened 2025-07-15
add wrapper around our custom binding to parse (unsupported) parameters


psteinroe:
now that we have a custom lexer, we can simply build a preprocessor for libpg_query to replace unsupported tokens each as parameters.

list of parameters to support (incomplete):
- `:name`
- ...

reteps:
Requesting `$name` as well 🙏

psteinroe:
> Requesting `$name` as well 🙏

tracking in #474

PR #473, opened 2025-08-12
fix: positional params

### LINKED ISSUE #405: Disable pg: Invalid statement: syntax error at or near ":" [syntax] rule
# Improve documentation

## Link 

Add a link to the page which needs improvement (if relevant)

## Describe the problem

Unsure how to disable this rule

## Describe the improvement

A way to disable the rule

## Additional context

Add any other context or screenshots that help clarify your question.

psteinroe:
pre-processes the query string to replace named params like `@name` with positional params like `$1`, because only the latter is supported by the postgres parser.

not super happy with the way it is used for the type checker but also did not want to put that function in another crate.

also removed a `println!` statement leftover from a previous pr of me.

also fixed a bug in the lexer that tokenized a cast as a named parameter.

fixes #405, #454


crates/pgt_workspace/src/workspace/server.tests.rs
```diff
@@ -277,3 +277,57 @@ async fn test_dedupe_diagnostics(test_db: PgPool) {
...
+    let setup_sql = r"
+      create table users (
+          id serial primary key,
+          name text not null,
+          email text not null
+      );
+    ";
+    test_db.execute(setup_sql).await.expect("setup sql failed");
+
+    let content = r#"select * from users where id = @one and name = :two and email = :'three';"#;
```
juleswritescode, line 311: `:'three'` – i want to meet the person who writes syntax like this haha

crates/pgt_workspace/src/workspace/server/pg_query.rs
```diff
@@ -87,10 +88,60 @@ impl PgQueryStore {
...
+    result
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
 
+    #[test]
+    fn test_convert_to_positional_params() {
+        let input = "select * from users where id = @one and name = :two and email = :'three';";
```
juleswritescode, line 147: what about duplicate identifier names?: 
```sql
select * from users 
where first_name = @one 
and starts_with(email, @one) 
and created_at > @two;
```

juleswritescode (changes requested):
just wondering about parameter reusage?

crates/pgt_workspace/src/workspace/server/pg_query.rs
psteinroe: good catch! fixed it

crates/pgt_workspace/src/workspace/server.tests.rs
psteinroe: haha yeah I did for some time two years ago :D

PR #475, opened 2025-08-12
fix: add support for named param type `$name`

### LINKED ISSUE #474: add support for `$name` in the lexer

psteinroe:

closes #474 

from https://github.com/supabase-community/postgres-language-server/issues/454#issuecomment-3177986456

### DIFF
--- a/crates/pgt_tokenizer/src/lib.rs
+++ b/crates/pgt_tokenizer/src/lib.rs
@@ -186,9 +186,29 @@ impl Cursor<'_> {
             '$' => {
                 // Dollar quoted strings
                 if is_ident_start(self.first()) || self.first() == '$' {
-                    self.dollar_quoted_string()
+                    // Get the start sequence of the dollar quote, i.e., 'foo' in $foo$hello$foo$
+                    // if ident does not continue and there is no terminating dollar
+                    // sign, we have a positional param `$name`
+                    let mut start = vec![];
+                    loop {
+                        match self.first() {
+                            '$' => {
+                                self.bump();
+                                break self.dollar_quoted_string(start);
+                            }
+                            c if is_ident_cont(c) => {
+                                self.bump();
+                                start.push(c);
+                            }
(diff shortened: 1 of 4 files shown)


crates/pgt_tokenizer/src/lib.rs
```diff
@@ -186,9 +186,29 @@ impl Cursor<'_> {
...
-                    self.dollar_quoted_string()
+                    // Get the start sequence of the dollar quote, i.e., 'foo' in $foo$hello$foo$
+                    // if ident does not continue and there is no terminating dollar
+                    // sign, we have a positional param `$name`
+                    let mut start = vec![];
+                    loop {
+                        match self.first() {
+                            '$' => {
+                                self.bump();
+                                break self.dollar_quoted_string(start);
```
juleswritescode, line 197:
```suggestion
                                return self.dollar_quoted_string(start);
```

crates/pgt_tokenizer/src/lib.rs
```diff
@@ -186,9 +186,29 @@ impl Cursor<'_> {
...
+                            }
+                            c if is_ident_cont(c) => {
+                                self.bump();
+                                start.push(c);
+                            }
+                            _ => {
+                                break TokenKind::NamedParam {
```
juleswritescode, line 204:
```suggestion
                                return TokenKind::NamedParam {
```

crates/pgt_tokenizer/src/lib.rs
juleswritescode: hat mich kurz irritiert, weil der breaked value ja auch aus der obrigen fn returned wird

crates/pgt_tokenizer/src/lib.rs
psteinroe: this would actually behave differently. we need break, and the value is then returned from the loop. with return, we would return the value from the function.

PR #484, opened 2025-08-30
feat: provide additional search_path patterns for typechecking 

juleswritescode:
Adds a `linter.searchPathPatterns` option where a user can add glob-patterns for additional schemas to be searched when typechecking.

Screenshot 2025-08-30 at 07 54 00

Screenshot 2025-08-30 at 07 54 13

Screenshot 2025-08-30 at 07 54 30

### DIFF
--- a/crates/pgt_schema_cache/Cargo.toml
+++ b/crates/pgt_schema_cache/Cargo.toml
@@ -23,7 +23,6 @@ sqlx.workspace            = true
 strum                     = { workspace = true }
 tokio.workspace           = true
 
-
 [dev-dependencies]
 pgt_test_utils.workspace = true
 
--- a/crates/pgt_schema_cache/Cargo.toml
+++ b/crates/pgt_schema_cache/Cargo.toml
@@ -15,7 +15,6 @@ version              = "0.0.0"
 anyhow.workspace          = true
 async-std                 = { version = "1.12.0" }
 futures-util              = "0.3.31"
-globset                   = "0.4.16"
 pgt_console.workspace     = true
 pgt_diagnostics.workspace = true
 serde.workspace           = true
@@ -23,7 +22,6 @@ serde_json.workspace      = true
 sqlx.workspace            = true
 strum                     = { workspace = true }
 tokio.workspace           = true
-itertools                 = "0.14.0"
 
 
 [dev-dependencies]
--- a/crates/pgt_schema_cache/Cargo.toml
+++ b/crates/pgt_schema_cache/Cargo.toml
@@ -15,6 +15,7 @@ version              = "0.0.0"
 anyhow.workspace          = true
 async-std                 = { version = "1.12.0" }
 futures-util              = "0.3.31"
+globset                   = "0.4.16"
 pgt_console.workspace     = true
 pgt_diagnostics.workspace = true
 serde.workspace           = true
@@ -22,6 +23,8 @@ serde_json.workspace      = true
 sqlx.workspace            = true
 strum                     = { workspace = true }
 tokio.workspace           = true
+itertools                 = "0.14.0"
+
 
 [dev-dependencies]
 pgt_test_utils.workspace = true
--- a/crates/pgt_schema_cache/src/schema_cache.rs
+++ b/crates/pgt_schema_cache/src/schema_cache.rs
@@ -89,7 +89,6 @@ impl SchemaCache {
             .filter(|t| t.name == name && schema.is_none_or(|s| s == t.schema.as_str()))
             .collect()
     }
-
 }
 
 pub trait SchemaCacheItem {
--- a/crates/pgt_schema_cache/src/schema_cache.rs
+++ b/crates/pgt_schema_cache/src/schema_cache.rs
@@ -1,4 +1,3 @@
-use globset::Glob;
 use sqlx::postgres::PgPool;
 
 use crate::columns::Column;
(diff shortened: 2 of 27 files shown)


crates/pgt_schema_cache/Cargo.toml
```diff
@@ -23,6 +23,7 @@ sqlx.workspace            = true
 strum                     = { workspace = true }
 tokio.workspace           = true
 
+
```
juleswritescode:
```suggestion
```

crates/pgt_schema_cache/src/schema_cache.rs
```diff
@@ -89,6 +89,7 @@ impl SchemaCache {
             .filter(|t| t.name == name && schema.is_none_or(|s| s == t.schema.as_str()))
             .collect()
     }
+
```
juleswritescode:
```suggestion
```

postgrestools.jsonc
```diff
@@ -12,7 +12,8 @@
     "enabled": true,
     "rules": {
       "recommended": true
-    }
+    },
+    "searchPathPatterns": ["pr*"]
```
juleswritescode:
```suggestion
    }
```

crates/pgt_typecheck/src/lib.rs
```diff
@@ -17,6 +20,9 @@ pub struct TypecheckParams<'a> {
     pub tree: &'a tree_sitter::Tree,
     pub schema_cache: &'a pgt_schema_cache::SchemaCache,
     pub identifiers: Vec<TypedIdentifier>,
+    /// Set of glob patterns that will be matched against the schemas in the database.
+    /// Each matching schema will be added to the search_path for the typecheck.
+    pub search_path_patterns: Vec<String>,
```
psteinroe, line 25: can't we compile the glob patterns already here?

juleswritescode: i tried at first to compile them when the settings were instantiated, but there are two problems:

1) the settings can change during a workspace session
2) and even worse, the schema_cache can be updated during a workspace session – since it's loaded lazily, there's no real spot where we can compile them

juleswritescode:
> is linter the right place for this? Or rather db settings?

Hmm had it there at first :D The reason I used the linter in the end was that the search_path only applies to the type checking, and to nothing else DB related. for example, if we use the EXECUTE STATEMENT feature, it'll not use those search paths. 

I also fear that there might be a misunderstanding that you "configure the db to use the search_path".

But I don't have a strong opinion on this. What do you prefer?

crates/pgt_typecheck/src/lib.rs
juleswritescode: we can however compile them in the workspace and pass them to the typecheck. Maybe that's better

juleswritescode: hmm for me it feels best to leave it as is, since the typecheck needs to decide what to do with the search patterns – e.g. the workspace doesn't know that the typecheck needs schema_names for a `SET search_path` statement.

psteinroe: makes sense!

psteinroe:
> > is linter the right place for this? Or rather db settings?
> 
> 
> 
> Hmm had it there at first :D The reason I used the linter in the end was that the search_path only applies to the type checking, and to nothing else DB related. for example, if we use the EXECUTE STATEMENT feature, it'll not use those search paths. 
> 
> 
> 
> I also fear that there might be a misunderstanding that you "configure the db to use the search_path".
> 
> 
> 
> But I don't have a strong opinion on this. What do you prefer?

mhh what do you think of adding a new section into he settings for typechecking? That way, we can also add disable etc. linting and type checking are two different things after all. and the lint settings are used to configure individual rules.

juleswritescode:
> > > is linter the right place for this? Or rather db settings?
> > 
> > 
> > Hmm had it there at first :D The reason I used the linter in the end was that the search_path only applies to the type checking, and to nothing else DB related. for example, if we use the EXECUTE STATEMENT feature, it'll not use those search paths.
> > I also fear that there might be a misunderstanding that you "configure the db to use the search_path".
> > But I don't have a strong opinion on this. What do you prefer?
> 
> mhh what do you think of adding a new section into he settings for typechecking? That way, we can also add disable etc. linting and type checking are two different things after all. and the lint settings are used to configure individual rules.

I think that's a good idea! I'll do it in this PR

juleswritescode:
done!

crates/pgt_workspace/src/workspace/server.rs
```diff
@@ -455,6 +455,7 @@ impl Workspace for WorkspaceServer {
             let path_clone = params.path.clone();
             let schema_cache = self.schema_cache.load(pool.clone())?;
             let input = doc.iter(TypecheckDiagnosticsMapper).collect::<Vec<_>>();
+            let search_path_patterns = settings.typecheck.search_path.clone();
```
psteinroe, line 458: nit: why not passing the entire config struct?

psteinroe (approved):
nice!

crates/pgt_workspace/src/workspace/server.rs
juleswritescode: because of a circular dependency: workspace depends on typecheck, typecheck depends on workspace::TypecheckSettings

PR #492, opened 2025-09-02
chore(flake): fix flake.nix

### LINKED ISSUE #481: Project's flake is incomplete.
# Chore

## Describe the chore

The flake provided on the repo does not produce a viable development environment.

## Additional context

The project's README file mentions that one can use `nix develop` to set up a suitable environment. However, this environment is incomplete: trying to compile the codebase will fail for several reasons:
- Compilation fails at `pgt_query` with a `bindgen` error, because `libclang` is not included in the environment.
- `bindgen` also fails to find headers from the standard library unless `LIBCLANG_PATH` is properly set.[^1]
- Compilation of `tikv-jemalloc-sys` fails. This one is trickier, the issue comes from `nixpkgs` and their environment, because it adds the `-Wall` flag during compilation.[^2][^3]

In addition, the development environment is missing some packages: `docker`, `docker-compose` and `docker-buildx`. I think the latter may not be necessary, but Docker will issue a warning if it isn't present.

Finally, on the topic of unused packages: I'm not sure all the settings included on the flake are necessary, but I just started reading through the codebase, so I thought better not to remove anything.

I changed the `flake.nix` to address the above issues, and can confirm that it's working on NixOS. I'd be more than happy to open a PR if you want.

Cheers!

---
## Relevant issues
[^1]: Unset `LIBCLANG_PATH` [issue](https://github.com/NixOS/nixpkgs/issues/52447)
[^2]: `tikv-jemalloc-sys` build fails on newer clang packages from `nixpkgs`: [issue](https://github.com/tikv/jemallocator/issues/108).
[^3]: Some versions of GCC on `nixpkgs` have unwanted flags: [issue](https://github.com/NixOS/nixpkgs/issues/18995).

Several have been reported multiple times, I added the ones that offered a solution.

2coffee2crab:

Chore: fixes some issues with the project's flake.nix file.

Compilation fails on the shell produced by the flake.

## What is the new behavior?

Compilation works as intended.

## Additional context

Closes #481.

I suspect that some settings may not be necessary, but I'd rather be conservative and not remove anything just yet.


psteinroe (approved):
thanks for the contribution!

PR #499, opened 2025-09-05
feat(hover): hover on roles, improve policy parser, change hover-layout

juleswritescode:
Screenshot 2025-09-05 at 13 13 42

Screenshot 2025-09-05 at 13 13 36

Screenshot 2025-09-05 at 13 13 59


crates/pgt_schema_cache/src/schema_cache.rs
```diff
@@ -100,6 +100,15 @@ impl SchemaCache {
             .filter(|f| f.name == name && schema.is_none_or(|s| s == f.schema.as_str()))
             .collect()
     }
+
+    pub fn find_roles(&self, name: &str) -> Vec<&Role> {
+        self.roles.iter().filter(|r| r.name == name).collect()
+    }
+}
+
+pub enum CommentSearch {
```
psteinroe: I think this is unused?

psteinroe (approved):
really like the structure in `pgt_hover`!

crates/pgt_schema_cache/src/schema_cache.rs
juleswritescode: ah yes you're right. I first tried loading all comments into schema_cache and searching from there, but then I realized there are helper functions for the shdesc table

PR #500, opened 2025-09-05
fix(hover): hover multiple statements, hover table in policy/revoke/grant

juleswritescode:


crates/pgt_workspace/src/workspace/server.rs
```diff
@@ -742,7 +747,7 @@ impl Workspace for WorkspaceServer {
             .next()
         {
             Some((stmt_id, range, ts_tree, maybe_ast)) => {
-                let position_in_stmt = params.position + range.start();
+                let position_in_stmt = params.position - range.start();
```
juleswritescode, line 750: this led to errors when we had multiple statements in a file. Only the first could be hovered

psteinroe: this might also fix #503?

juleswritescode: Yes I believe so!

PR #502, opened 2025-09-05
feat: autocomplete with single `"` 

juleswritescode:
- [x] fix policy completions – seem to be broken

given `select | from public.users "u"` or `select | from public.users u`…

- autocomplete with single `"` -> results in `"u"."email"` or `"email"` 
- autocomplete within double `""` -> results in `"u"."email"` or `"email"`
- recognize aliases regardless of whether they are quoted or not: 
    - `"u".|` and `u.|` will both suggest `email`
    - `"u"."|` and `u."|` and `"u"."|"` and `u."|"` will all suggest `"email"` 
    - *This does not yet work for tables -> separate PR*
- existing behavior without quotes is unaffected


psteinroe (approved):
`has_uneven_quotes` - this is getting ridiculous 😂

Always Sunny Reaction GIF

PR #508, opened 2025-09-10
feat: port remaining squawk rules

### LINKED ISSUE #131: Lint
The lint crate just needs more rules! Its design is very much inspired by [Squawk](https://github.com/sbdchd/squawk), and we should at least add some rules from there. There is also an [open issue to publish squawk as a crate](https://github.com/sbdchd/squawk/issues/333), which would allow us to use it directly instead of copying over its code…

- [ ] add more rules (add your request below)
- [ ] regularly check issue about publishing Squawk as a crate

psteinroe:
will have claude do the work over night

closes #131

will port https://github.com/kaaveland/eugene/blob/main/eugene/src/lints/rules.rs afterwards

also added a new `agentic/` dir and two new `just` commands to standardise as well as track the prompts for such work.

UPDATE: 
- refactored some of the rules to make them easier to consume.
- added a bunch of tests
- refactored the test framework to allow multiple statements per file (required to test some rules)
- refactored the test framework to require explicit declaring each expected diagnostic
- migrated `prefer jsonb` from eugene too. the remaining ones are either covered be squawk or require a bit more updates to the file context. will do them in a follow-up. see full comparison here: https://github.com/supabase-community/postgres-language-server/discussions/305#discussioncomment-14424302


crates/pgt_analyser/src/lint/safety/prefer_big_int.rs
```diff
@@ -0,0 +1,127 @@
...
+                let is_small_int = matches!(
+                    type_name_lower.as_str(),
+                    "smallint"
+                        | "integer"
+                        | "int2"
+                        | "int4"
+                        | "serial"
+                        | "serial2"
+                        | "serial4"
+                        | "smallserial"
```
juleswritescode, line 108: hab jetzt hier nur mal nen kleinen stichproben test gemacht, die anderen rules hab ich nicht angeschaut.

durch diesen check würde der error fälschlicherweise jetzt auch für integer, int4, serial, serial4 reported – vielleicht sollten wir claud sagen, dass er auch noch tests für false positives machen soll

psteinroe: das ist auch so gewollt. alle integer Varianten sollten durch bigint ersetzt werden. hier das equivalent in squawk:

https://github.com/sbdchd/squawk/blob/master/crates/squawk_linter/src/rules/prefer_bigint_over_int.rs

https://github.com/sbdchd/squawk/blob/master/crates/squawk_linter/src/rules/prefer_bigint_over_smallint.rs

aber ja, schaue die Woche nochmal alles durch und packe tests dazu

justfile
```diff
@@ -152,3 +152,22 @@ quick-modify:
 # just show-logs | bunyan
 show-logs:
     tail -f $(ls $PGT_LOG_PATH/server.log.* | sort -t- -k2,2 -k3,3 -k4,4 | tail -n 1)
+
+agentic name:
+    unset ANTHROPIC_API_KEY && claude --dangerously-skip-permissions -p "please read agentic/{{name}}.md and follow the instructions closely"
```
juleswritescode: not sure we should commit this into an open source repo – seems like a vulnerability for prompt injection :)
We could pass a --settings `agentic/{{name}}.settings.json` file ? 

Also, I'm curious: Why do you unset the API key here? To force somebody to login?

juleswritescode (approved):

justfile
psteinroe: how would that be an attack vector?

I am removing the env because otherwise it doesn't pick up the Max subscription but the API key (which is usage based)

juleswritescode: > I am removing the env because otherwise it doesn't pick up the Max subscription but the API key (which is usage based)

ah i see! that makes sense.

> how would that be an attack vector?

for example if a malicious contributor adds something into a build script that replaces the contents of your `agentic/` file. If you then run the agentic file, claude might do [stuff you don't expect](https://www.youtube.com/watch?v=Rxn_X04lg88&t=150s) – with GH permissions, claude might even push unreviewd commits to master etc

If we review everything closely, that won't be an issue, but I could see that someone contributes a `C` file or something that we just gloss over

Of course, this would still happen if you run `--dangerously-skip-permissions` via your own terminal, but then at least nobody would know that we use that param 😇

Your call!

psteinroe: good point, I just commented them out for now with a node 'use at your own risk'

PR #509, opened 2025-09-11
feat(completions): correctly infer (quoted) schema for tables & functions

juleswritescode:
- `select "auth"."| ;` will now correctly expand to `"auth"."uid"()` 
- `select "auth"."|" ;` as well
-  `select "auth".|` expands to `"auth".uid()` 

also: before, if there were quotes around a schema, we wouldn't find the possible tables/functions. now we do 🥳


psteinroe (approved):
nice!

PR #513, opened 2025-09-12
fix: context-aware named param conversion

### LINKED ISSUE #510: Support graphile migrate placeholders
Graphile migrate uses placeholder for some variables. They are compiled to the matching env variables at runtime. 

Image

It would be great if the pg-lsp would support that, or at least allow to disable some rule that causes this error.

psteinroe:
we were blindly converting named parameters to positional parameters. but the latter is only valid as a literal, and not as an identifier.

statements like

```sql
grant usage on schema public, app_public, app_hidden to :DB_ROLE;
```

are not valid when converted to

```sql
grant usage on schema public, app_public, app_hidden to $1;
```

i went a bit back and forth on this and decided the easiest way to fix this is to convert to identifiers like `a` if the previous non-trivia token is one of a set list of tokens. We will probably have a bunch of edge cases here but fixing them should be as easy as adding a keyword to the list.

now, we convert to

```sql
grant usage on schema public, app_public, app_hidden to a;
```

closes #510

### DIFF
--- a/crates/pgt_lexer/src/lexer.rs
+++ b/crates/pgt_lexer/src/lexer.rs
@@ -143,7 +143,7 @@ impl<'a> Lexer<'a> {
                         }
                         _ => {}
                     };
-                    SyntaxKind::POSITIONAL_PARAM
+                    SyntaxKind::NAMED_PARAM
                 }
                 pgt_tokenizer::TokenKind::QuotedIdent { terminated } => {
                     if !terminated {
--- a/crates/pgt_workspace/src/workspace/server/pg_query.rs
+++ b/crates/pgt_workspace/src/workspace/server/pg_query.rs
@@ -3,9 +3,9 @@ use std::num::NonZeroUsize;
 use std::sync::{Arc, LazyLock, Mutex};
 
 use lru::LruCache;
+use pgt_lexer::lex;
 use pgt_query_ext::diagnostics::*;
 use pgt_text_size::TextRange;
-use pgt_tokenizer::tokenize;
 use regex::Regex;
 
 use super::statement_identifier::StatementId;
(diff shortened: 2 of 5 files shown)


crates/pgt_workspace/src/workspace/server/pg_query.rs
```diff
@@ -146,17 +179,45 @@ pub fn convert_to_positional_params(text: &str) -> String {
...
+/// Generates a deterministic identifier based on the given index.
+fn deterministic_identifier(idx: usize) -> String {
+    let iteration = idx / ALPHABET.len();
+    let pos = idx % ALPHABET.len();
+
+    format!(
+        "{}{}",
+        ALPHABET[pos],
+        if iteration > 0 {
+            deterministic_identifier(iteration - 1)
```
juleswritescode, line 201: i love me a good recursion

crates/pgt_lexer/src/lexer.rs
```diff
@@ -143,7 +143,7 @@ impl<'a> Lexer<'a> {
                         }
                         _ => {}
                     };
-                    SyntaxKind::POSITIONAL_PARAM
+                    SyntaxKind::NAMED_PARAM
```
juleswritescode, line 146: das war n bug nehm ich an?

juleswritescode (approved):
nice! 

So the typechecker doesn't properly check whether a role exists or not? 
I would've expected it to complain about the "missing role `a`"

psteinroe:
> nice! 
> 
> 
> 
> So the typechecker doesn't properly check whether a role exists? 
> 
> I would've expected it to complain about the "missing role `a`" 

The typechecker just runs on DML statements and such replacement are less of an issue there. but at some point someone will report it too I guess. If that happens, we can find another solution for it. 😅👍🏼

crates/pgt_workspace/src/workspace/server/pg_query.rs
psteinroe: das kam sogar ohne Claude 😂

crates/pgt_lexer/src/lexer.rs
psteinroe: Ja, allerdings kein wichtiger für die Statement Splitter Logik deshalb kam es nie raus

PR #527, opened 2025-09-19
chore: replace tree-sitter submodule with internal crate

juleswritescode:
This also made it possible to upgrade to treesitter `0.25.X` – I only had to adjust two lines in `pgt_treesitter/context/mod.rs`. :)


crates/pgt_treesitter_grammar/src/scanner.c
```diff
@@ -0,0 +1,188 @@
+#include "tree_sitter/parser.h"
```
juleswritescode, line 1: this is not autogenerated, but copied from the derek stride repo.
You can add context-aware parsing to `treesitter` via the `scanner.c` file.
This is only necessary for parsing dollar-quoted strings `$tag$ select 1; $tag$`

crates/pgt_treesitter_grammar/grammar.js
```diff
@@ -0,0 +1,3405 @@
+/**
```
juleswritescode, line 1: also copied from the derek stride repo, I just added JSDoc types and fixed some linting errors

psteinroe (approved):
Awesome!

PR #530, opened 2025-09-19
refactor(completions): remove custom parse for policies, use tree-sitter

juleswritescode:


justfile
```diff
@@ -11,13 +11,15 @@ alias qm := quick-modify
 # Installs the tools needed to develop
 install-tools:
 	cargo install cargo-binstall
+	cargo install tree-sitter-cli
```
juleswritescode, line 14: binstall not supported :( https://github.com/tree-sitter/tree-sitter/issues/4351

PR #534, opened 2025-09-21
fix: add missing column in group by expression

aokiji:
when trying to use the lsp server autocompletion didnt work and when activating logs found messages such as:

> ERROR pgt_lsp::session error=Database error: column "c.relkind" must appear in the GROUP BY clause or be used in an aggregate function

upon adding c.relkind into the group by expression autocomplete started working as expected, the issue seems to be using c.relkind in the select statement without using it in the group by expression

I'm using postgres 12 to test it


psteinroe:
> when trying to use the lsp server autocompletion didnt work and when activating logs found messages such as:
> 
> > ERROR pgt_lsp::session error=Database error: column "c.relkind" must appear in the GROUP BY clause or be used in an aggregate function
> 
> upon adding c.relkind into the group by expression autocomplete started working as expected, the issue seems to be using c.relkind in the select statement without using it in the group by expression
> 
> I'm using postgres 12 to test it

weird that sqlx prepare did not throw - thanks!

PR #541, opened 2025-09-27
feat(typecheck): also replace longer type defaults

juleswritescode:
Before, we would only show type errors if the length of the default-type-value was <= than the length of the param-name it belonged to. 

Now we have an API that tracks positions in text replacements even if they grow the original text, so we can show such errors: 

Screenshot 2025-09-27 at 14 26 07


crates/pgt_typecheck/src/typed_identifier.rs
```diff
@@ -325,15 +314,62 @@ mod tests {
...
+        let replacement = super::apply_identifiers(identifiers, &schema_cache, &tree, input);
 
-        assert!(valid_pos);
         assert_eq!(
-            sql_out,
+            replacement.text(),
             // the numeric parameters are filled with 0;
-            // all values of the enums are longer than `NULL`, so we use `NULL` instead
-            "select 0      + 0                           + 0  + 0                   + 0               + NULL     "
+            "select 0      + 0                           + 0  + 0                   + 0               + 'critical'"
```
juleswritescode, line 317: not sure if the change from `NULL` to `'critical'` is desired?

crates/pgt_text_size/src/text_range_replacement.rs
```diff
@@ -0,0 +1,353 @@
...
+                        marker.original_range.end()
+                            + marker.registered_offset_at_start
+                            + marker.range_size_difference,
+                    );
+
+                    total_offset += usize::from(marker.range_size_difference);
+                }
+                AdjustmentDirection::Shortened => {
+                    unimplemented!(
+                        "So far, we've only ever lenghtened TextRanges. Consider filling up your range with spaces"
```
juleswritescode:
```suggestion
                        "So far, we've only ever lengthened TextRanges. Consider filling up your range with spaces"
```

crates/pgt_text_size/src/text_range_replacement.rs
```diff
@@ -0,0 +1,353 @@
...
+                        marker.original_range.start() + marker.registered_offset_at_start,
+                        marker.original_range.end()
+                            + marker.registered_offset_at_start
+                            + marker.range_size_difference,
+                    );
+
+                    total_offset += usize::from(marker.range_size_difference);
+                }
+                AdjustmentDirection::Shortened => {
+                    unimplemented!(
```
psteinroe, line 132: nit: why not supporting this? Would be more consistent

crates/pgt_typecheck/src/typed_identifier.rs
psteinroe: yes, makes sense after your change

psteinroe (approved):
Beautiful!

crates/pgt_text_size/src/text_range_replacement.rs
juleswritescode: just because it's not needed right now – I'll do it in a follow up :)

PR #543, opened 2025-10-01
chore: dual publish

psteinroe:
this pr set ups dual publish to the `@postgres-language-server` namespace. also:
- after seeing `pgls` in the wild I decided we should name the binary `postgres-language-server`
- we now copy the `README.md` into each package when releasing 
- added a deprecation notice to the `postgestools` packages' readme

idea is that we 
- wait for the npm namespace 
- publish a new release 
- update all integrations / extensions / docs / env vars / config file name to point to the new package / binaries / name
- publish a new release, mark the `postgrestools` npm packages as deprecated
- wait for a few weeks, then remove dual publishing

not sure if that is a good idea. maybe we also need to support both the old and the new config name, env vars etc at the same time, but not sure if its worth the effort? what do you think?


packages/@postgrestools/postgrestools/scripts/generate-packages.mjs
```diff
@@ -219,6 +219,35 @@ function copySchemaToNativePackage(platform, arch) {
 	fs.chmodSync(schemaTarget, 0o666);
 }
 
+function copyReadmeToPackage(packagePath) {
+	const packageRoot = resolve(PACKAGES_POSTGRESTOOLS_ROOT, packagePath);
+	const readmeSrc = resolve(POSTGRESTOOLS_ROOT, "README.md");
```
juleswritescode, line 224: do we need to adjust the `README.md`? there are probably some `postgrestools` mentions in there, right?

.github/workflows/publish.reusable.yml
```diff
@@ -33,15 +33,23 @@ jobs:
             exit 1
           fi
 
+      - name: Generate Packages (deprecated)
+        id: generate-packages-deprecated
+        run: node packages/@postgres-language-server/cli/scripts/generate-packages.mjs
```
juleswritescode: shouldn't this point at `postgrestools`?

.github/workflows/publish.reusable.yml
```diff
@@ -50,7 +58,16 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }} #
 
-      - name: Publish npm packages as latest
+      - name: Publish npm packages as nightly (deprecated)
+        if: inputs.is-prerelease == 'true'
+        run: |
+          for package in packages/@postgres-language-server/*; do
```
juleswritescode, line 64: shouldn't this point at `postgrestools`?

.github/workflows/publish.reusable.yml
```diff
@@ -66,3 +83,20 @@ jobs:
           done
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish npm packages as latest
+        if: inputs.is-prerelease != 'true'
```
juleswritescode, line 88: I think the nightly upload isn't setup for the new name

packages/@postgres-language-server/backend-jsonrpc/src/transport.ts
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	/**
+	 * Send a request to the remote server
+	 *
+	 * @param method Name of the remote method to call
+	 * @param params Parameters object the remote method should be called with
+	 * @return Promise resolving with the value returned by the remote method, or rejecting with an RPC error if the remote call failed
+	 */
+	// biome-ignore lint/suspicious/noExplicitAny: if i change it to Promise<unknown> typescript breaks
```
juleswritescode, line 129: hier ist noch n `biome` command drin

packages/@postgrestools/postgrestools/scripts/generate-packages.mjs
psteinroe: nope, only in the link to the vscode extension

packages/@postgres-language-server/backend-jsonrpc/src/transport.ts
psteinroe: because we are using biome to lint this ts file and we need to ignore the lint :D

.github/workflows/publish.reusable.yml
psteinroe: I am blind, thanks!

.github/workflows/publish.reusable.yml
psteinroe: I am blind, thanks!

.github/workflows/publish.reusable.yml
psteinroe: do we even need to setup anything for it? and we aren't using it anyway, so I thought just copy it for now to not loose it, but is-prerelease is always false

juleswritescode: i'm good with removing it. we just needed it for the first release, I think :)

packages/@postgres-language-server/backend-jsonrpc/src/transport.ts
juleswritescode: > I am blind, thanks!

me too apparently 😆

packages/@postgrestools/postgrestools/scripts/generate-packages.mjs
juleswritescode: nice!

.github/workflows/publish.reusable.yml
```diff
@@ -33,15 +33,23 @@ jobs:
             exit 1
           fi
 
+      - name: Generate Packages (deprecated)
+        id: generate-packages-deprecated
+        run: node packages/@postgres-language-server/postgrestools/scripts/generate-packages.mjs
```
juleswritescode:
```suggestion
        run: node packages/@postgrestools/postgrestools/scripts/generate-packages.mjs
```

juleswritescode (approved):
i just committed a suggestion here, please make sure that checks out 👍🏻 
rest LGTM!

PR #544, opened 2025-10-01
chore: dual publish

psteinroe:
attempt to support both names for env vars, package names and config files


psteinroe:
doing this in chunks now

PR #547, opened 2025-10-03
refactor: type replacement tracker supports shortening ranges

juleswritescode:


crates/pgt_typecheck/tests/diagnostics.rs
```diff
@@ -74,7 +112,66 @@ async fn invalid_column(pool: PgPool) {
...
+                schema: None,
+                name: "uuid".to_string(),
+                is_array: false,
+            },
+        }],
+    }
+    .test()
+    .await;
+
+    /* NOTE: The replaced type default value is *longer* than the param name. */
```
juleswritescode:
```suggestion
    /* NOTE: The replaced type default value is *shorter* than the param name. */
```

Issue #516, opened 2025-09-17
thread `tokio-runtime-worker` has overflowed its stack


diraneyya:
<img width="852" height="551" alt="Image" src="https://github.com/user-attachments/assets/328cb23e-ae6c-4f18-8e5c-2e58c0790f8d" />

Currently updated `plpgsql_check` to 2.8.2, while using the most recent version of the language server (0.13). The VS Code updated automatically.

I am not sure if this is coming from the extension or the language server but the crash happens when I edit a certain function in an extension SQL file.

psteinroe:
hey, thanks for reporting - that's a new one! 

is it reproducible? If so, would you mind sharing the file and location that causes the crash?

diraneyya:
Hello, thank you for the quick response, as usual.

Currently I am not able to reproduce it, but it was crashing consistently at the time.

Since then I have been trying to trace another obscure error https://github.com/supabase-community/postgres-language-server/issues/515 which led me to this thread https://github.com/okbob/plpgsql_check/issues/178, in which the maintainer of `plpgsql_check` has recommended to add `plpgsql_check` early in the `shared_preload_libraries` setting.

Since I made this change and restarted my PostgreSQL server. I have not experienced it. But if I did I will get back here and report how to reproduce it reliably if it happened again.

juleswritescode:
I'm pretty sure this is related to #554, I can reproduce it when typing out the case statement with the format function that's in the screenshot above.

<img width="400" height="174" alt="Image" src="https://github.com/user-attachments/assets/15c9ed95-d1eb-45e3-9085-ea3556794496" />

We have the same treesitter structure where a child is smaller than its parent: 

<img width="344" height="80" alt="Image" src="https://github.com/user-attachments/assets/715d0069-25f7-4e05-89ca-767696025bf6" />

So I'm confident the PR #555 will fix it 🫶

PR #555, opened 2025-10-12
fix: infinite recursion with SQL params

### LINKED ISSUE #554: Stack Overflow when TS Child Node is smaller than Parent Node
The problem happens when the treesitter parses parses a child node to be smaller than the parent node, and the cursor position is *not* on the child node:

```
@i|d -- parent node with cursor
@    -- child node
```

Image

juleswritescode:
Fixes #554 and maybe #516


crates/pgt_treesitter/src/context/mod.rs
```diff
@@ -1214,4 +1216,45 @@ mod tests {
             _ => unreachable!(),
         }
     }
+
+    #[test]
+    fn does_not_overflow_callstack_on_smaller_treesitter_child() {
+        //  Instead of autocompleting "FROM", we'll assume that the user
+        // is selecting a certain column name, such as `frozen_account`.
```
juleswritescode:
```suggestion
```

PR #556, opened 2025-10-12
fix(splitter): improve EOF handling

### LINKED ISSUE #540: pgt_statement_splitter error
# Bug report

## Describe the bug

2025-07-25 12:09:46.314 [info] Encountered an unexpected error
Source Location: crates\pgt_statement_splitter\src\splitter.rs:115:14
Thread Name: tokio-runtime-worker
Message: lexed should have non-trivia eof token

2025-07-25 12:09:46.597 [info] [Error - 12:09:46 PM] PostgresTools language server closed
2025-07-25 12:09:46.597 [info] [Error - 12:09:46 PM] Server process exited with code 0.

## To Reproduce

Open and edit some SQL file in VS Code

## System information

- OS: Windows 11
- VS Code Version: 1.102.2
Commit: c306e94f98122556ca081f527b466015e1bc37b0
Date: 2025-07-22T12:15:48.520Z
Electron: 35.6.0
ElectronBuildId: 11847422
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Windows_NT x64 10.0.26100

juleswritescode:
Currently, if you type `select case ` in a file, the LSP will crash.

The reason is that we forgot to handle the EOF token in the `case()` function in the splitter.

I feel that, since we deal with incomplete statements all the time, it's better to bubble up an Err if we reach EOF:

- it's an unrecoverable error and should immediately stop splitting
- this way, we'll never forget adding the EOF to any statement again

But it's a bit invasive, so WDYT? 

(Pretty sure this fixes #540, it's the same kind of error)


crates/pgt_statement_splitter/src/splitter/common.rs
```diff
@@ -8,117 +10,126 @@ use super::{
     dml::{cte, delete, insert, select, update},
 };
 
-pub fn source(p: &mut Splitter) {
+#[derive(Debug)]
+pub struct SplitterException;
+
+impl std::fmt::Display for SplitterException {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // todo
```
juleswritescode:
```suggestion
```

crates/pgt_statement_splitter/src/splitter/common.rs
```diff
@@ -128,43 +138,44 @@ pub(crate) fn plpgsql_command(p: &mut Splitter) {
...
 
-pub(crate) fn unknown(p: &mut Splitter, exclude: &[SyntaxKind]) {
+pub(crate) fn unknown(p: &mut Splitter, exclude: &[SyntaxKind]) -> Result<(), ReachedEOFException> {
     loop {
         match p.current() {
             SyntaxKind::SEMICOLON => {
-                p.advance();
-                break;
-            }
-            SyntaxKind::EOF => {
```
juleswritescode, line 156: note: we do not need to handle EOF below anymore

PR #557, opened 2025-10-12
feat: hover on types

juleswritescode:


crates/pgt_treesitter/src/context/mod.rs
```diff
@@ -205,6 +205,8 @@ impl<'a> TreesitterContext<'a> {
             ctx.gather_info_from_ts_queries();
         }
 
+        println!("{:#?}", ctx);
```
psteinroe: oversight?

juleswritescode: yes, great catch, thanks :)

juleswritescode: removed

PR #562, opened 2025-10-23
[1] refactor: remove custome revoke parser

juleswritescode:
complexity be gone! we can just rely on treesitter grammar for revoke statements.


psteinroe (approved):
its iconic how much more sense this makes compared what we had before!

Issue #302, opened 2025-04-04
Add support for running a command to get the database connection string in `postgrestools.jsonc`


psteinroe:
### Discussed in https://github.com/supabase-community/postgres-language-server/discussions/292 and https://github.com/supabase-community/postgres-language-server/discussions/472

<div type='discussions-op-text'>

<sup>Originally posted by **remorses** April  1, 2025</sup>
I want to use the Supabase database directly instead of a local database, to do that you could add support for a config option `dbCommand` to get the database uri via stdout

```jsonc
{
  "$schema": "./docs/schemas/latest/schema.json",
  // get the database uri using dotenv-cli npm package
  "dbCommand": "dotenv -- bash -c 'echo $DATABASE_URL'",
  // get database uri using doppler
  "dbCommand": "doppler secrets get --plain DATABASE_URL",
}
```

This allows to get the database url using tools like Doppler or dotenv
</div>

additional feedback: loading the url values from an environment variable (see second linked discussion)

willruggiano:
I would also love the ability to use InitializeParams and/or the workspace/didChangeConfiguration notification to configure this stuff, rather than be limited to a configuration file. The latter could enable switching database connections. Both options would then allow the client to determine the db parameters however they wanted, e.g. something along the lines of (in Neovim with lspconfig);

```lua
require("lspconfig").postgres_lsp.setup {
  settings = {
    postgres_lsp = {
      db = load_from_env()
    },
  }
}
```

```lua
function switch_database_connection(client --[[@type vim.lsp.Client]])
  client.notify(
    "workspace/didChangeConfiguration",
    {
      settings = { ... as above ... }
    }
  )
end
```

PR #573, opened 2025-10-26
feat(config): add connectionString

### LINKED ISSUE #575: allow connecting via `connectionString`
### Discussed in https://github.com/supabase-community/postgres-language-server/discussions/572

<sup>Originally posted by **gnakic** October 25, 2025</sup>
Either using a dev branch of my db instance or a local docker container, I need to add the following to jsonc config
`?sslmode=require&channel_binding=require`
but not sure where exactly. Does anyone know? Adding those as query params to either `database` or `host` does not work.</div>

psteinroe:
adds `connectionString` to the database config. if provided, it is preferred over the individual `host`, `port`, etc. settings.

note that the `ConnectionKey` still checks only based on the basic infos. This could potentially lead to an issue if the same workspace instance accepts multiple connections to the same database. if one is e.g. adding different `sslmode` settings via the connection sring, it would be re-used even if another just has the same `host` + `port` + `user` + `database`. I do not think we need to distinguish here since that case should not really happen unless you host this as a service that is shared among many people.

fixes #575


psteinroe:
> love it, maybe this is a good opportunity to also finish #302 ?

yeah, it could be as simple as adding a ENV to the cli option

PR #574, opened 2025-10-26
refactor(cli): simplify abstractions

psteinroe:
this PR refactors the `cli` crate. it currently has a bunch of issues that stem from its original implementation in biome, where file/stdin execution is the main use case.

- a `CommandRunner` trait that tries to abstract a few common code blocks away.
- a `Dummy` traversal mode which essentially means "do not traverse".
- reports are tightly coupled to execution. we also need to report output of commands that do not traverse (e.g. `dblint`)
- `execute_mode` does everything: stdin handling + traversal + reporting + exit codes. it cannot be re-used by commands that do not traverse

all of these issues cause my brain to have a hard time understanding what is going on. This PR refactors the crate to gear it more towards our use-case. I spent a few cycles designing the refactor, and at least for me it makes much more sense now.

I also added a dummy `dblint` command that will be implemented in follow-ups. right now, it prints an empty report.

Below is a richer summary of the issues described above.

### 1. CommandRunner Trait Forces All Commands Through Same Pipeline
```rust
pub(crate) trait CommandRunner: Sized {
    fn run(&mut self, session: CliSession, cli_options: &CliOptions) -> Result<()> {
        // Forces: config loading → VCS setup → file traversal → reporting
        let (execution, paths) = self.configure_workspace(fs, console, workspace, cli_options)?;
        execute_mode(execution, session, cli_options, paths)
    }
}
```

**Issues:**
- Only `Check` command uses this trait
- Commands like `version`, `clean`, `init` are ad-hoc functions
- No way for commands to skip parts they don't need (e.g., dblint needs config but not traversal)
- Forces unnecessary complexity on simple commands

### 2. TraversalMode::Dummy Exists For Wrong Reasons
```rust
pub enum TraversalMode {
    Dummy,  // Only exists because commands are forced through traversal
    Check { stdin: Option<Stdin>, vcs_targeted: VcsTargeted },
}
```

Commands that don't need traversal shouldn't have a "dummy" mode - they just shouldn't traverse.

### 3. Execution Struct Conflates Concerns
```rust
pub struct Execution {
    report_mode: ReportMode,      // How to report (Terminal/GitHub/GitLab)
    traversal_mode: TraversalMode, // How to process files
    max_diagnostics: u32,
}
```

**Problem:** Bundles processing config with reporting config. Commands that don't traverse (like `dblint`) still need reporting but don't need traversal config.

### 4. execute_mode() is Monolithic
```rust
pub fn execute_mode(execution: Execution, session: CliSession, ...) -> Result<()> {
    // Does: stdin handling + traversal + reporting + exit codes
    // Can't be reused by commands that don't traverse
}
```

### 5. Scattered and Inconsistent Structure
- `commands/` has mix of trait impls and functions
- `execute/` has traversal logic tightly coupled to reporting
- No clear separation of workspace setup, execution, and reporting
- Commands repeat boilerplate for config loading, setup, reporting

### DIFF
--- a/crates/pgt_cli/src/commands/check.rs
+++ b/crates/pgt_cli/src/commands/check.rs
@@ -11,7 +11,6 @@ use pgt_fs::FileSystem;
 use pgt_workspace::DynRef;
 use std::ffi::OsString;
 
-#[allow(dead_code)]
 pub struct CheckArgs {
     pub configuration: Option<PartialConfiguration>,
     pub paths: Vec<OsString>,
@@ -19,8 +18,6 @@ pub struct CheckArgs {
     pub staged: bool,
     pub changed: bool,
     pub since: Option<String>,
-    pub apply: bool,
-    pub apply_unsafe: bool,
 }
 
 pub fn check(
--- a/crates/pgt_cli/src/execute/process_file/workspace_file.rs
+++ b/crates/pgt_cli/src/execute/process_file/workspace_file.rs
@@ -9,8 +9,11 @@ use std::path::{Path, PathBuf};
 /// Small wrapper that holds information and operations around the current processed file
 pub(crate) struct WorkspaceFile<'ctx, 'app> {
     guard: FileGuard<'app, dyn Workspace + 'ctx>,
+    /// File handle for the underlying filesystem entry, if backed by a real file.
+    /// Not present for stdin execution where content is provided as a temporary buffer.
+    /// Currently unused but will be needed when autofix/write operations are implemented.
     #[allow(dead_code)]
-    file: Option<Box<dyn File>>, // only present when backed by a real fs entry
+    file: Option<Box<dyn File>>,
     pub(crate) path: PathBuf,
 }
 
--- a/crates/pgt_cli/src/reporter/gitlab.rs
+++ b/crates/pgt_cli/src/reporter/gitlab.rs
@@ -139,12 +139,19 @@ impl Display for GitLabDiagnostics<'_> {
     }
 }
 
+/// An entry in the GitLab Code Quality report.
+/// See https://docs.gitlab.com/ee/ci/testing/code_quality.html#implement-a-custom-tool
 #[derive(Serialize)]
 pub struct GitLabDiagnostic<'a> {
+    /// A description of the code quality violation.
     description: String,
+    /// A unique name representing the static analysis check that emitted this issue.
     check_name: &'a str,
+    /// A unique fingerprint to identify the code quality violation. For example, an MD5 hash.
     fingerprint: String,
+    /// A severity string (can be info, minor, major, critical, or blocker).
     severity: &'a str,
+    /// The location where the code quality violation occurred.
     location: Location,
 }
 
--- a/crates/pgt_cli/src/execute/walk.rs
+++ b/crates/pgt_cli/src/execute/walk.rs
@@ -161,8 +161,10 @@ fn traverse_inputs(
 
 struct DiagnosticsPrinter<'ctx> {
     _config: &'ctx ExecutionConfig,
+    /// The maximum number of diagnostics the console thread is allowed to print
     max_diagnostics: u32,
     remaining_diagnostics: AtomicU32,
+    /// Count of diagnostics that exceeded max_diagnostics and weren't printed
     not_printed_diagnostics: AtomicU32,
     printed_diagnostics: AtomicU32,
     total_skipped_suggested_fixes: AtomicU32,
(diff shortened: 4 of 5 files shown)


PLAN.md
juleswritescode, line 1: I suppose this shouldn't be committed?

crates/pgt_cli/src/reporter/gitlab.rs
```diff
@@ -155,19 +139,12 @@ impl Display for GitLabDiagnostics<'_> {
     }
 }
 
-/// An entry in the GitLab Code Quality report.
```
juleswritescode, line 158: these seem like useful comments?

crates/pgt_cli/src/commands/dblint.rs
```diff
@@ -0,0 +1,18 @@
...
+pub fn dblint(
+    mut session: CliSession,
+    cli_options: &CliOptions,
+    cli_configuration: Option<PartialConfiguration>,
+) -> Result<(), CliDiagnostic> {
+    let configuration = session.prepare_with_config(cli_options, cli_configuration)?;
+    session.setup_workspace(configuration, VcsIntegration::Disabled)?;
+
+    // TODO: Implement actual dblint logic here
+    let report = Report::new(vec![], std::time::Duration::new(0, 0), 0, None);
```
juleswritescode, line 15: this is clearly a peak MVP

crates/pgt_cli/src/commands/check.rs
```diff
@@ -1,76 +1,141 @@
...
+#[allow(dead_code)]
+pub struct CheckArgs {
+    pub configuration: Option<PartialConfiguration>,
+    pub paths: Vec<OsString>,
+    pub stdin_file_path: Option<String>,
+    pub staged: bool,
+    pub changed: bool,
+    pub since: Option<String>,
+    pub apply: bool,
+    pub apply_unsafe: bool,
```
juleswritescode: these seem to be dead, do we need them?
if not, let's remove the unnecessary complexity 👍🏻

crates/pgt_cli/src/reporter/mod.rs
```diff
@@ -3,61 +3,147 @@ pub(crate) mod gitlab;
...
+            traversal,
+        }
+    }
 }
 
-/// When using this trait, the type that implements this trait is the one that will **write** the data, ideally inside a buffer
-pub trait ReporterVisitor {
-    /// Writes the summary in the underling writer
-    fn report_summary(
+pub trait ReportWriter {
```
juleswritescode, line 96: I love that the trait was called "Visitor" before, with a command specifying that it would **write**

crates/pgt_cli/src/execute/traverse.rs
```diff
@@ -187,69 +160,31 @@ fn traverse_inputs(
 // struct DiagnosticsReporter<'ctx> {}
 
 struct DiagnosticsPrinter<'ctx> {
-    ///  Execution of the traversal
-    #[allow(dead_code)]
-    execution: &'ctx Execution,
-    /// The maximum number of diagnostics the console thread is allowed to print
+    _config: &'ctx ExecutionConfig,
     max_diagnostics: u32,
-    /// The approximate number of diagnostics the console will print before
```
juleswritescode, line 195: sanity check: is the comment regarding "skipped diagnostics" counter useful?

crates/pgt_cli/src/execute/process_file/workspace_file.rs
```diff
@@ -10,7 +10,7 @@ use std::path::{Path, PathBuf};
 pub(crate) struct WorkspaceFile<'ctx, 'app> {
     guard: FileGuard<'app, dyn Workspace + 'ctx>,
     #[allow(dead_code)]
-    file: Box<dyn File>,
+    file: Option<Box<dyn File>>, // only present when backed by a real fs entry
```
juleswritescode: this is still confusing to me:

1. apparently `file` is never read – should we delete it?
2. if we want to keep it, how can a `WorkspaceFile` not be backed by a real fs entry? In case it's a temporary buffer I suppose – but let's add some clarification in a comment 👍🏻

juleswritescode (approved):
much simpler! love it.

crates/pgt_cli/src/execute/process_file/workspace_file.rs
psteinroe: its for `stdin` execution - which we currently do not have a real use for since we never write in our commands. this will become used once we have autofix.

PR #579, opened 2025-10-28
[2] refactor: use more specific tree-sitter references and identifiers

juleswritescode:
- introduces a couple of new leaf-identifier nodes: `role_identifier`, `table_identifier` and so on
- introduces a couple of new referenves next to the old `object_reference`, e.g. `table_reference` or `function_reference`. 
- adds queryable tags to the fields of the references and refactors the `tree-sitter` queries to use them

Notice that we can only use precise leaf-identifiers if we have a fully qualified `*_reference`:

In an incomplete statement, if a user writes `select us| from some_table;`, the `us` could be either a schema, a table, an alias, or a column. The user is still writing the query; we don't know what he'll type next!
 
Even if he writes `select something.us|`, `something` might refer to a table, alias or schema; `us` might refer to a table or column.

Only when he writes `select something.users.email` do we know which part is a schema, which a table, and which a column.

The tags on the `*_reference` should help us decide which data types to consider for completion/hover.

---

In the next PRs, I want to store relevant parent nodes and field names. Right now, we're traversing the tree alot, for example to find out if a node has a certain tag or if the ancestory history matches a template. We can make this easier and faster by collecting the traversed nodes/fields in some kind of data structure and checking against that.

### DIFF
--- a/crates/pgls_treesitter_grammar/tests/snapshots/grammar_tests__test_2_sql3.snap
+++ b/crates/pgls_treesitter_grammar/tests/snapshots/grammar_tests__test_2_sql3.snap
@@ -1,0 +1,26 @@
+---
+source: crates/pgls_treesitter_grammar/tests/grammar_tests.rs
+expression: writer
+---
+update auth.users set auth.users.email = 'my@mail.com';
+-----------------------
+program [0..55] 'update auth.users set auth.users.email = 'my@mail.com';'
+  statement [0..54] 'update auth.users set auth.users.email = 'my@mail.com''
+    update [0..54] 'update auth.users set auth.users.email = 'my@mail.com''
+      keyword_update [0..6] 'update'
+      relation [7..17] 'auth.users'
+        table_reference [7..17] 'auth.users'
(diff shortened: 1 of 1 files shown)


crates/pgls_treesitter_grammar/tests/grammar_tests.rs
```diff
@@ -0,0 +1,314 @@
...
+          keyword_set [18..21] 'set'
+          assignment [22..43] 'email = 'my@mail.com''
+            column_reference [22..27] 'email'
+              any_identifier [22..27] 'email'
+            = [28..29] '='
+            literal [30..43] ''my@mail.com''
+      ; [43..44] ';'
+    ");
+
+    assert_snapshot!(printed_tree(sql2), @r"
```
psteinroe: warum nicht file snapshots?

juleswritescode: yes, viel besser!

die ersten waren noch kurz, da fand ich inline besser

PR #593, opened 2025-11-01
chore(diagnostics): add database object as diagnostics location

psteinroe:
prep for dblint

### DIFF
--- a/packages/@postgres-language-server/backend-jsonrpc/src/workspace.ts
+++ b/packages/@postgres-language-server/backend-jsonrpc/src/workspace.ts
@@ -139,7 +139,11 @@ export type Advice =
 /**
  * Represents the resource a diagnostic is associated with.
  */
-export type Resource_for_String = "argv" | "memory" | { file: string };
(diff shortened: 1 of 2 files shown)


crates/pgls_diagnostics/src/location.rs
```diff
@@ -4,6 +4,28 @@ use std::fmt::Debug;
 use std::ops::Range;
 use std::{borrow::Borrow, ops::Deref};
 
+/// Represents a database object (table, function, etc.)
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct DatabaseObject<'a> {
+    /// Optional schema name
+    pub schema: Option<&'a str>,
+    /// Object name (required)
+    pub name: &'a str,
```
juleswritescode, line 13: just a thought: We'll probably refer to columns here as well? what will the name look like in that case, "table.column"?

psteinroe: Yes, I would suggest we can add them when they are needed

PR #594, opened 2025-11-01
chore(splinter): initial commit

psteinroe:
adds a new `pgls_splinter` crate with the basic integration.

its not hooked up yet.

pr includes

- the query execution
- the transformation into splinter diagnostics
- a codegen tool for the diagnostics categories that extracts them from the sql file
- a build file that downloads and parses the splinter.sql file from the repo. its cached and committed. The build script is just a helper when we want to "upgrade". we only have to change the SHA in build.rs for that. 
- a check that splinter is only executed if supabase roles are in the target database

the query currently crashes when executed against a non-supabase database because of the calls to `pg_catalog.has_table_privilege('anon', c.oid, 'SELECT')` in some of the rules. for now, we skip splinter if the roles are not there. I opened an issue to ask for permission to add a flag to each rule so it can be used outside of supabase too: https://github.com/supabase/splinter/issues/135. I would rather not make it our own just yet, but thats certainly a simple fix if things take too long.

UPDATE: manually split up the query into supabase-only and generic rules in a follow-up.

follow up prs will include:
- refactoring of the analyser codegen so we can reuse it for other lint tools to generate config, docs etc
- integration of the new DatabaseLocation attribute of the diagnostics struct (pr is open)

### DIFF
--- a/crates/pgls_completions/src/providers/roles.rs
+++ b/crates/pgls_completions/src/providers/roles.rs
@@ -54,6 +54,10 @@ mod tests {
                 crate::CompletionItemKind::Role,
             ),
             CompletionAssertion::LabelAndKind("owner".into(), crate::CompletionItemKind::Role),
+            CompletionAssertion::LabelAndKind(
+                "service_role".into(),
+                crate::CompletionItemKind::Role,
+            ),
             CompletionAssertion::LabelAndKind("test_login".into(), crate::CompletionItemKind::Role),
             CompletionAssertion::LabelAndKind(
                 "test_nologin".into(),


crates/pgls_splinter/TODO.md
juleswritescode, line 1: personal note?

crates/pgls_splinter/build.rs
```diff
@@ -0,0 +1,109 @@
+use std::env;
+use std::fs;
+use std::path::Path;
+
+// Update this commit SHA to pull in a new version of splinter.sql
+const SPLINTER_COMMIT_SHA: &str = "27ea2ece65464213e466cd969cc61b6940d16219";
```
juleswritescode, line 6: do we need the sha in the vendor file as well?

crates/pgls_splinter/src/query.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub cache_key: String,
+}
+
+pub async fn load_splinter_results(pool: &PgPool) -> Result<Vec<SplinterQueryResult>, sqlx::Error> {
+    let mut tx = pool.begin().await?;
+
+    // this is done by the splinter.sql file normally, but we remove it so that sqlx can work with
+    // the file properly.
+    sqlx::query("set local search_path = ''")
+        .execute(&mut *tx)
```
juleswritescode: the pool might use a different connection for this statement than for the coming query, so the search_path might not be properly set in some cases 🤔 

In `plgs_typecheck/lib`, we're using a connection and not a pool for the same reason – maybe we should do that here, too?

crates/pgls_splinter/tests/diagnostics.rs
```diff
@@ -0,0 +1,260 @@
...
+struct TestSetup<'a> {
+    name: &'a str,
+    setup: &'a str,
+    test_db: &'a PgPool,
+}
+
+impl TestSetup<'_> {
+    async fn test(self) {
+        // Create Supabase-specific roles that splinter expects
+        for role in ["anon", "authenticated", "service_role"] {
```
juleswritescode: we need to set up roles via a sqlx migration file, because setting them up via a single sqlx test database will set them up at the instance level and therefore for *all* sqlx test databases (test pollution).

There's a `0001_setup-roles.sql` file in the repo, we should set up those roles there 👍🏻

PR #595, opened 2025-11-02
[3] refactor: improve & simplify tree-sitter context-tracking

juleswritescode:
- introduced a new scope_tracker that makes context tracking easier and will allow us to discard context when we enter subqueries.
- `ctx.node_under_cursor` does not need to be an Option; we should always have a tree.root_node
- deleted a couple of obsolete helper methods on the `TreesitterContext`

Next up, I'll make sure we open new scopes for CTEs, subqueries etc.
Also, I think the whole WrappingClause and WrappingNodes are unnecessary indirections – we can just as easily check the tree-sitter node kind `&str`s.

### DIFF
--- a/crates/pgls_completions/src/test_helper.rs
+++ b/crates/pgls_completions/src/test_helper.rs
@@ -199,7 +199,5 @@ pub(crate) async fn assert_no_complete_results(query: &str, setup: Option<&str>,
     let params = get_test_params(&tree, &cache, query.into());
     let items = complete(params);
 
-    println!("{items:#?}");
-
     assert_eq!(items.len(), 0)
 }


crates/pgls_completions/src/providers/columns.rs
```diff
@@ -717,18 +716,6 @@ mod tests {
             &pool,
         )
         .await;
-
-        // no completions in the values list!
-        assert_no_complete_results(
```
juleswritescode, line 722: deleted this for now; I think we should have completions in the values list, but making sure they are the right suggestions is another issue.

crates/pgls_treesitter/src/context/base_parser.rs
juleswritescode, line 1: was unused since the latest revoke/grant parser refactors 👍🏻

PR #597, opened 2025-11-04
refractor(splinter): allow non-supabase rules to run always

psteinroe:
manually splits the `splinter.sql` file into generic and supabase-only rules during download in `build.rs`.

based on the existence of the roles, run either one or both.

also improved the url handling - no manual fix needed anymore.


crates/pgls_splinter/src/lib.rs
```diff
@@ -32,17 +32,19 @@ async fn check_required_roles(conn: &PgPool) -> Result<bool, sqlx::Error> {
...
-        return Ok(Vec::new());
-    }
+    let mut all_results = Vec::new();
+
+    let generic_results = query::load_generic_splinter_results(params.conn).await?;
+    all_results.extend(generic_results);
 
-    let results = query::load_splinter_results(params.conn).await?;
+    // Only run Supabase-specific rules if the required roles exist
+    let has_supabase_roles = check_supabase_roles(params.conn).await?;
```
juleswritescode, line 41: nice! curious: did the supabase team confirm that this check is the best way to identify a supabase database?

crates/pgls_splinter/build.rs
```diff
@@ -107,3 +128,39 @@ fn add_not_null_markers(content: &str) -> String {
 
     result
 }
+
+fn split_queries(content: &str) -> (String, String) {
```
juleswritescode, line 163: as I understand it, rules are per default put into the generic category.
what happens if supabase adds a new rule to the spliter that's supabase-specific?
that might break linting in case a table isn't defined for non-supabase databases, right?
should we throw during build if there's an unknown rule in the splinter?

crates/pgls_splinter/src/lib.rs
psteinroe: nope, but for squawk this is what is required to exist for the rules query to not crash :D

crates/pgls_splinter/build.rs
psteinroe: great catch!

PR #598, opened 2025-11-04
refactor: config codegen

psteinroe:
rewires the pgls_configuration crate alongside its codegen xtask as a preparation for new lint tools (`splinter` and `pglinter`). its all the same as before, I just moved things in the config crate around, and the xtask is geared towards tools vs just the analyser. common primitives are now in `/rules/`, and we generate one dir per tool (`linter`, `splinter`, `pglinter`) etc. 

the codegen refactoring was initially written by me but completed by Claude. Everything outside of xtask/ was done by hand or is re-generated.

## TODO (in follow-up prs)
- Rename `pgls_analyser` crate to `pgls_linter`
- Create `pgls_pglinter`, `pgls_splinter` and `pgls_assists` and integrate them with `pgls_analyse` to allow a `RuleVisitor` to collect the metadata
- Simplify common patterns between linter/splinter/pglinter - I think we can cleanup `Analyser` a bit.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "postgres_lsp",


crates/pgls_configuration/src/rules/selector.rs
```diff
@@ -0,0 +1,92 @@
+use pgls_analyse::RuleFilter;
+
+use std::str::FromStr;
+
+use crate::{Rules, linter::RuleGroup};
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub enum RuleSelector {
```
psteinroe, line 8: this will be made more generic in a future pr

psteinroe:
> Looks good!  
> 
> 
> 
> The way I understand it is: "rules" are an abstraction layer over lints, actions, transformations.
> 
> 
> 
> Then the "lints" will come from many tools: splinter, pglinter, and so on.
> 
> 
> 
> What's still throwing me off a bit is that *our* lint rules (mostly squawk, eugene, typecheck?) are called "linter" right now. Maybe we can somehow make it more clear that `linter` is one of *many* linters by naming it "internal" or something?
> 
> 
> 
> 
> 
> 

The idea is that "rules" are the abstraction for not only lints, actions and transformations, but also for diagnostics emitted from other tools such as splinter. 

and the linter is the "source code rules engine", and it emits "normal" lints. we will then add another crate for assists and transformations if we need them. 

The other tools (eg splinter) are also declaring "rules" and emit diagnostics for rules, but are not declared as "linter".

But I agree that it's not 100% perfect. Open to suggestions, but I have troubles finding a name for the "source code linter" that is not "linter".

juleswritescode:
Alright, let's merge it!

PR #599, opened 2025-11-08
test(completions): add test suite for more coverage

juleswritescode:
Writing every last variation of an SQL statement felt unmaintable, and it was annoying to work with the completions tests.

The new suite can be used to write large statements, check completions for every node, and to verify completion results more quickly.


crates/pgls_completions/src/relevance/filtering.rs
```diff
@@ -324,7 +324,7 @@ impl CompletionFilter<'_> {
                         WrappingClause::DropRole | WrappingClause::AlterRole => true,
 
                         WrappingClause::SetStatement => ctx
-                            .before_cursor_matches_kind(&["keyword_role", "keyword_authorization"]),
+                            .before_cursor_matches_kind(&["eyword_role", "keyword_authorization"]),
```
juleswritescode:
```suggestion
                            .before_cursor_matches_kind(&["keyword_role", "keyword_authorization"]),
```

crates/pgls_completions/src/snapshots/pgls_completions__test_helper__prefers_not_mentioned_columns.snap
juleswritescode, line 1: now that I'm reading this, it's probably cool to add the setup to the snapshot as well, for context.

juleswritescode: done!

juleswritescode:
@psteinroe finally reviewable – my local `bun` was outdated, so ci failed because of a dirty `bun.lock` 😢

PR #618, opened 2025-12-14
chore(splinter): integrate with codegen

psteinroe:
the actual changes are just within the codegen task. we now generate config and rule metadata from the sql too.

the flow is now:
1. generate splinter rule metadata (`rules/` directory) from splinter sql
2. use a `Visitor` to collect rule metadata from the registry that loads the generated rule metadata when generating analyser code such as config and docs

This allows us to reuse all existing `analyse` infrastructure outside of the `analyser` / ast-linter.


psteinroe:
> very cool — the sql is just taken from SB, right?

yes!

PR #625, opened 2025-12-19
Feat/keyword completion

juleswritescode:


juleswritescode:
closed in favor of the approach in #629

PR #631, opened 2025-12-28
feat: add dblint with splinter

psteinroe:
as simple as it gets for now.

will wire up pglinter next and then check if we can refactor linters to share even more

fixed the generated documentation too, which makes up for most of the changes.


crates/pgls_workspace/src/workspace/server/analyser.rs
```diff
@@ -138,18 +161,119 @@ impl RegistryVisitor for LintVisitor<'_, '_> {
...
+
+            self.enabled_rules.extend(enabled_rules);
+
+            let disabled_rules = self
+                .settings
+                .as_splinter_rules()
+                .map(|rules| rules.as_disabled_rules())
+                .unwrap_or_default();
+            self.disabled_rules.extend(disabled_rules);
+        }
```
juleswritescode, line 212: this seems very familiar, is this an opportunity to dedupe some code?

PR #632, opened 2025-12-29
feat(dblint): add pglinter

psteinroe:
adds support for [`pglinter`](https://github.com/pmpetit/pglinter) and integrate it with `dblint`. It reuses the existing infrastructure, and 99% of the code changes are generated.

adapted the `Dockerfile` to also install `pglinter`. Will take around 10 minutes now to build, but its all cached after that.

the initial metadata is generated from the downloaded repo. a follow-up pr will automate this too.

pglinter works in a weird way where the user defines thresholds for certain rules. e.g.

```
---
source: crates/pgls_pglinter/tests/diagnostics.rs
expression: content
snapshot_kind: text
---
Category: pglinter/base/howManyObjectsWithUppercase
Severity: Warning
Message: Count number of objects with uppercase in name or in columns.
Advices:
Count number of objects with uppercase in name or in columns.
[Info] Rule: B005
How to fix:
[Info]   1. Do not use uppercase for any database objects
```

I would much rather see the actual objects that have an uppercase in name...

Opened an issue to check in with the maintainer: https://github.com/pmpetit/pglinter/issues/83

### DIFF
--- a/.github/actions/setup-postgres/action.yml
+++ b/.github/actions/setup-postgres/action.yml
@@ -78,9 +78,9 @@ runs:
         # Initialize pgrx for the installed PostgreSQL version
         cargo pgrx init --pg${PG_VERSION} $(which pg_config)
 
-        # Clone and build pglinter (clone to /tmp, use feat/83/violation_list for rule_messages)
+        # Clone and build pglinter (requires v1.1.0+ for get_violations API + rule_messages table)
         cd /tmp
-        git clone -b feat/83/violation_list https://github.com/pmpetit/pglinter.git
+        git clone --depth 1 https://github.com/pmpetit/pglinter.git
         cd pglinter
 
         # Install using pgrx
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,9 +19,9 @@ RUN apt-get update && \
     cargo install cargo-pgrx --version 0.16.1 --locked && \
     # Initialize pgrx for PostgreSQL 15
     cargo pgrx init --pg15 $(which pg_config) && \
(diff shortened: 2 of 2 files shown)


.github/actions/setup-postgres/action.yml
```diff
@@ -53,30 +54,77 @@ runs:
...
+        cd /tmp
+        git clone https://github.com/pmpetit/pglinter.git
+        cd pglinter
+
+        # Install using pgrx
+        cargo pgrx install --pg-config $(which pg_config) --release
+
+        # Verify installation
+        echo "Extension control files:"
+        ls -la "$(pg_config --sharedir)/extension/" | grep pglinter || echo "No pglinter found"
```
juleswritescode, line 91: should we fail early here? right now this would be cachted after creating extension `plpgsql_check`, if I'm not mistaken

crates/pgls_pglinter/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    let disabled_in_extension = match cache {
+        Some(c) => c.disabled_rules.clone(),
+        None => cache::get_disabled_rules(params.conn).await?,
+    };
+
+    // Check for mismatches and collect runnable rules
+    let mut runnable_rules = Vec::new();
+    for rule_code in &enabled_rules {
+        if disabled_in_extension.contains(rule_code) {
+            results.push(PglinterDiagnostic::rule_disabled_in_extension(rule_code));
```
juleswritescode, line 122: love it

juleswritescode (approved):
sweet!

PR #655, opened 2026-01-24
2/11: docs(grammar): add grammar guidelines

juleswritescode:
## Summary
- Add documentation for grammar contributions
- Explain identifier types, partial grammars, and `@end` markers

This is PR 2/11 in the keyword completion feature stack.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3112,6 +3112,7 @@ name = "pgls_treesitter_grammar"
 version = "0.0.0"
 dependencies = [
  "cc",
+ "criterion",
  "insta",
  "pgls_test_utils",
  "tree-sitter",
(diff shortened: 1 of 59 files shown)


crates/pgls_treesitter_grammar/GRAMMAR_GUIDELINES.md
```diff
@@ -0,0 +1,254 @@
...
+    $.column_identifier,
+    optional($._alias),
+    $.keyword_from,
+    field("end", $.table_reference)
+  );
+```
+
+Now, if the user types `select email as e|`, the resulting looks like `keyword_select column_identifier keyword_as any_identifier(@end)`, and the select statement is prematurely considered completed.
+
+However, we could hypothetically make the `$.table_reference` a private `$._table_reference` and put an `"end"` node in there. The clause would still complete at the right spot.
```
psteinroe, line 249: Maybe a note about what private vs public fields actually are?

juleswritescode: I think that should be well-known; it's one of the standard tree-sitter syntaxes: https://tree-sitter.github.io/tree-sitter/creating-parsers/3-writing-the-grammar.html#hiding-rules

PR #656, opened 2026-01-24
3/11: feat(grammar): add benchmarks

juleswritescode:
## Summary
- Add criterion dependency
- Add parsing benchmarks
- Add lookahead iterator benchmark

This is PR 3/11 in the keyword completion feature stack.


juleswritescode:
> Doesn't criterion support writing benchmark results to a file vs adding comments?

it per default writes benchmarks to the `/target` directory.

I think the recommended approach isn't to store any results, tbh, because it'll differ so much on varying machines.

What we could do is running benchmarks on the same gh-runner machine and look for regressions.

In this case, it was only meant as a POC that updating the tree-sitter tree once for every checkable keyword is feasible 👍🏻

PR #657, opened 2026-01-24
4/11: feat(treesitter): add helper module with navigation utilities

juleswritescode:
## Summary
- Add `goto_node_at_position`, `goto_previous_leaf`, `goto_closest_unfinished_parent_clause`
- Add `previous_sibling_completed`, `last_children_completed`
- Add `SINGLE_TOKEN_RULES` constant
- Export functions from lib.rs

This is PR 4/11 in the keyword completion feature stack.

### DIFF
--- a/crates/pgls_treesitter/src/helper.rs
+++ b/crates/pgls_treesitter/src/helper.rs
@@ -1,6 +1,6 @@
 use tree_sitter::{Node, Tree};
 
-pub static SINGLE_TOKEN_RULES: &[&'static str] = &[
+pub static SINGLE_TOKEN_RULES: &[&str] = &[
     "any_identifier",
     "column_identifier",
     "schema_identifier",
@@ -97,7 +97,7 @@ pub fn previous_sibling_completed(node: tree_sitter::Node) -> bool {
 
         is_finished && last_children_completed(prev)
     } else {
-        false
+        true
     }
 }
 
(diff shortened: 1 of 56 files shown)


crates/pgls_treesitter/src/helper.rs
```diff
@@ -0,0 +1,120 @@
+use tree_sitter::{Node, Tree};
+
+pub static SINGLE_TOKEN_RULES: &[&'static str] = &[
```
psteinroe: Can this be generated?

psteinroe: like what is this list / how is it defined?

juleswritescode: some treesitter nodes only ever have one child, such as `literal`:

```
program [0..12] 'select true;'
  statement [0..11] 'select true'
    select [0..11] 'select true'
      keyword_select [0..6] 'select'
      select_expression [7..11] 'true'
        term [7..11] 'true'
          literal [7..11] 'true'
            keyword_true [7..11] 'true'
  ; [11..12] ';'
```

So we're basically saying, "these token rules are always completed".

Alternatively, we could mark all children with an `@end` tag; I might do that later:

```
literal: ($) =>
      field("end", prec(
        2,
        choice(
          $._integer,
          $._decimal_number,
          $._literal_string,
          $._bit_string,
          $._string_casting,
          $.keyword_true,
          $.keyword_false,
          $.keyword_null
        )
      )),
```

PR #661, opened 2026-01-24
7/11: feat(completions): add keyword completion provider

juleswritescode:
## Summary
- Add `SqlKeyword` struct with keyword list
- Add `complete_keywords` function
- Add `Keyword` variant to `CompletionItemKind`
- Add filtering logic with speculative parsing
- Add keyword scoring (-10 to deprioritize)
- Update builder to pass shared tree for filtering

This is PR 7/11 in the keyword completion feature stack.

### DIFF
--- a/crates/pgls_completions/src/providers/keywords.rs
+++ b/crates/pgls_completions/src/providers/keywords.rs
@@ -535,7 +535,6 @@ mod tests {
     }
 
     #[sqlx::test]
-    #[ignore = "will be reintroduced after stacked keyword-completion PRs merge"]
     async fn completes_after_from_clause(pool: PgPool) {
         let setup = r#"
             create table public.users (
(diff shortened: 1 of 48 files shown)


crates/pgls_completions/src/providers/keywords.rs
```diff
@@ -0,0 +1,841 @@
...
+        self
+    }
+
+    const fn starts_statement(mut self) -> Self {
+        self.starts_statement = true;
+        self
+    }
+}
+
+pub static ALL_KEYWORDS: &[SqlKeyword] = &[
```
psteinroe, line 37: Can this be generated?

crates/pgls_completions/src/relevance/filtering.rs
```diff
@@ -409,6 +482,127 @@ impl CompletionFilter<'_> {
 
         Some(())
     }
+
+    /// Filters out keywords marked with `require_prefix` unless the user has started typing them.
```
psteinroe, line 486: Nice detail!

crates/pgls_completions/src/providers/keywords.rs
juleswritescode: I think we can generate the keywords, but we need some manual processing to mark them as "statement starters" or as "require prefix".

I'll look into that in a future PR.

PR #664, opened 2026-01-24
10/11: fix: helper edge case and typo

juleswritescode:
## Summary
- Fix `previous_sibling_completed` to return true when no sibling exists
- Fix typo "speciifcation" -> "specification"

This is PR 10/11 in the keyword completion feature stack.


crates/pgls_treesitter/src/context/mod.rs
```diff
@@ -384,7 +384,7 @@ impl<'a> TreesitterContext<'a> {
             "where" => Some(WrappingClause::Where),
             "update" => Some(WrappingClause::Update),
             "select" => Some(WrappingClause::Select),
-            "delete" => Some(WrappingClause::Delete),
+            "delete_statement" => Some(WrappingClause::Delete),
```
psteinroe, line 387: What's the difference between the names?

juleswritescode: "delete" doesn't even exist as a node 🤷🏻 

"delete_statement" should now be the parent of every statement starting with a delete keyword.

PR #674, opened 2026-02-07
feat: support standard postgres env vars for db connection

### LINKED ISSUE #302: Add support for running a command to get the database connection string in `postgrestools.jsonc`

psteinroe:
Support `DATABASE_URL`, `PGHOST`, `PGPORT`, `PGUSER`, `PGPASSWORD`, `PGDATABASE` env vars for configuring the database connection with highest priority.

we use bpaf(env()) for the cli, and support loading the db config from_env for the lsp.

These env var names follow the standard [libpq environment variables](https://www.postgresql.org/docs/current/libpq-envars.html) (`PG*`) and the widely adopted `DATABASE_URL` convention used by tools like `sqlx`, `diesel`, `prisma`, and `dotenv`.

### Example

```sh
DATABASE_URL=postgres://user:pass@localhost:5432/mydb postgres-language-server check file.sql
# or individual vars
PGHOST=localhost PGPORT=5432 PGUSER=user PGPASSWORD=pass PGDATABASE=mydb postgres-language-server check file.sql
```

### Precedence (highest to lowest)

1. Environment variables
2. CLI args / LSP client settings
3. Config file
4. Defaults

Closes #302

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -148,9 +148,6 @@ jobs:
       - name: Install JS dependencies
         run: bun install
 
-      - name: Restore lockfile after platform-specific install
-        run: git checkout bun.lock
-
       - name: Setup Just
         uses: extractions/setup-just@v3
 
--- a/bun.lock
+++ b/bun.lock
@@ -1,6 +1,5 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "postgres_lsp",
@@ -36,6 +35,14 @@
         "@postgres-language-server/cli-x86_64-linux-gnu": "<placeholder>",
         "@postgres-language-server/cli-x86_64-linux-musl": "<placeholder>",
         "@postgres-language-server/cli-x86_64-windows-msvc": "<placeholder>",
+      },
+    },
+    "packages/@postgres-language-server/wasm": {
+      "name": "@postgres-language-server/wasm",
+      "version": "0.0.0",
+      "devDependencies": {
+        "@playwright/test": "^1.50.0",
+        "typescript": "^5.0.0",
       },
     },
     "packages/@postgrestools/backend-jsonrpc": {
@@ -84,6 +91,8 @@
 
     "@biomejs/cli-win32-x64": ["@biomejs/cli-win32-x64@1.9.4", "", { "os": "win32", "cpu": "x64" }, "sha512-8Y5wMhVIPaWe6jw2H+KlEm4wP/f7EW3810ZLmDlrEEy5KvBsb9ECEfu/kMWD484ijfQ8+nIi0giMgu9g1UAuuA=="],
 
+    "@playwright/test": ["@playwright/test@1.58.2", "", { "dependencies": { "playwright": "1.58.2" }, "bin": { "playwright": "cli.js" } }, "sha512-akea+6bHYBBfA9uQqSYmlJXn61cTa+jbO87xVLCWbTqbWadRVmhxlXATaOjOgcBaWU4ePo0wB41KMFv3o35IXA=="],
+
     "@postgres-language-server/backend-jsonrpc": ["@postgres-language-server/backend-jsonrpc@workspace:packages/@postgres-language-server/backend-jsonrpc"],
 
     "@postgres-language-server/cli": ["@postgres-language-server/cli@workspace:packages/@postgres-language-server/cli"],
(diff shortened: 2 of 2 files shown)


crates/pgls_configuration/src/database.rs
```diff
@@ -11,29 +11,29 @@ use serde::{Deserialize, Serialize};
...
     pub port: u16,
 
     /// The username to connect to the database.
-    #[partial(bpaf(long("username")))]
+    #[partial(bpaf(env("PGUSER"), long("username")))]
     pub username: String,
 
     /// The password to connect to the database.
-    #[partial(bpaf(long("password")))]
+    #[partial(bpaf(env("PGPASSWORD"), long("password")))]
```
pacak: This might have an unexpected side effect of leaking password in the `--help` message. If that's not acceptable - I'd implement a separate function for obtaining the password and call it here with `bpaf(external(get_password))`.

This function would have a parser for password and a parser for env, hidden. Something like this:

```rust
fn get_password() -> impl Parser<String> {
    let pass = long("password")
           .help("The password to connect to the database")
           .argument("PASSWORD");
    let epass = env("PGPASSOWRD").argument("PASSWORD").hide();
    construct!([pass, epass])
}
```

psteinroe: thanks for noting this! super helpful, did not know that it leaks in `--help`.

crates/pgls_env/src/lib.rs
```diff
@@ -156,6 +180,24 @@ impl Display for PgLSEnv {
             }
         };
 
+        for var in [
+            &self.database_url,
+            &self.pghost,
+            &self.pgport,
+            &self.pguser,
+            &self.pgpassword,
```
juleswritescode: not sure if we want to log the password somewhere?

juleswritescode: we could alternatively wrap it in a [secret-string](https://crates.io/crates/secret-string)

crates/pgls_cli/src/lib.rs
```diff
@@ -197,6 +205,16 @@ impl<'app> CliSession<'app> {
             configuration.merge_with(cli_config);
         }
 
+        // Env vars take highest priority — merge last so they override everything.
+        if let Some(env_db) = pgls_configuration::database::PartialDatabaseConfiguration::from_env()
+        {
+            let env_config = PartialConfiguration {
+                db: Some(env_db),
+                ..Default::default()
```
juleswritescode, line 210: i'm a little bit confused about the ..default::default().. shouldnt we marge the configuration here? or are the defaults all "None"?

crates/pgls_configuration/src/database.rs
```diff
@@ -64,3 +70,39 @@ impl Default for DatabaseConfiguration {
...
+        }
+
+        Some(Self {
+            connection_string: database_url,
+            host: pghost,
+            port: pgport,
+            username: pguser,
+            password: pgpassword,
+            database: pgdatabase,
+            ..Default::default()
```
juleswritescode, line 127: ..Default::default() returns all `None`? 
just wondering if we need to be careful about not overwriting everything with default values?

crates/pgls_configuration/src/database.rs
pacak: Hmm... I'm wondering. Are there any scenarios where you want to see the value of the variable in the `--help` message? I could make it to tell the variable name and indicate if it is set or not. Or even just the name...

psteinroe: Mhh I think with just showing that it is set right now will cover the same use case that showing the actual value does today. Users could just echo it to reveal the value.

crates/pgls_env/src/lib.rs
psteinroe: i opted for the easy way now because using any wrapper type would mean adding biomes merge partial to our repo to support the type... but it might make sense to bring all the biome deps in-house anyways. will note it as a follow up.

crates/pgls_cli/src/lib.rs
psteinroe: i double checked and it does not overwrite

crates/pgls_configuration/src/database.rs
psteinroe: i double checked and it does not overwrite

crates/pgls_env/src/lib.rs
psteinroe: create a note to bring `biome_deserialise` in-house and replace all other `biome_` dependencies with ocx - their lib apis are much better and well maintained.

PR #686, opened 2026-02-18
feat(wasm): generate and export schema cache types 

psteinroe:


xtask/codegen/src/generate_bindings.rs
```diff
@@ -4,6 +4,10 @@ use pgls_workspace::workspace_types::{generate_type, methods, ModuleQueue};
 use xtask::{project_root, Mode, Result};
 
 pub fn generate_bindings(mode: Mode) -> Result<()> {
+    generate_workspace_bindings(&mode)
+}
+
+fn generate_workspace_bindings(mode: &Mode) -> Result<()> {
```
juleswritescode: I guess we could also use &mode below instead of adding indirection?

xtask/codegen/src/main.rs
```diff
@@ -37,6 +38,9 @@ fn main() -> Result<()> {
         TaskCommand::Pglinter => {
             generate_pglinter()?;
         }
+        TaskCommand::WasmSchemaTypes => {
```
juleswritescode: this isn't wasm-related per se, right? should we rename to `GenerateSchemaTypes` or something?

PR #688, opened 2026-02-22
chore: migrate from biome to oxc

psteinroe:
if we use it within the codebase, we should also use it as our formatter and linter

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -6,8 +6,8 @@
       "name": "postgres_lsp",
       "devDependencies": {
         "@types/bun": "latest",
-        "oxfmt": "^0.34.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.0.0",
   "private": true,
   "devDependencies": {
-    "@biomejs/biome": "1.9.4",
+    "oxlint": "^1.49.0",
+    "oxfmt": "^0.34.0",
```
juleswritescode: let's used fixed versions with all the supply chain shizzle 👍🏻 
```suggestion
    "oxlint": "1.49.0",
    "oxfmt": "0.34.0",
```

PR #689, opened 2026-02-22
feat(completions): finish grammar for COPY and INSERT statements

juleswritescode:


crates/pgls_completions/src/providers/keywords.rs
```diff
@@ -435,15 +435,15 @@ mod tests {
...
+                    "copy".into(),
                     crate::CompletionItemKind::Keyword,
                 ),
                 CompletionAssertion::LabelAndKind(
-                    "drop".into(),
+                    "create".into(),
                     crate::CompletionItemKind::Keyword,
                 ),
                 CompletionAssertion::LabelAndKind(
-                    "insert".into(),
```
juleswritescode, line 446: note: we lost "insert" here!

 apparently treesitter can only handle so many states at once, so with a completely fresh statement (`|`), treesitter now suggests `COPY`, but `INSERT` was kicked out.

This will happen more frequently and I don't want to work around that, but it should be fine since the keyword still shows up when you type `i|`: 

Image

juleswritescode:
> nice! question: would it make sense to publish the grammar as a tree-sitter-postgres variant? how well would it work for syntax highlighting?

we can try :) it's much more specific than the original postgres treesitter, which makes it more customizable but it's also more effort to implement.

never tried it though!

PR #690, opened 2026-02-23
fix: clean pending transport requests and add leak-check xtask probes

psteinroe:
## Summary
Fixes retained-memory growth in `SocketTransport::request` by ensuring pending request entries are removed on all error paths (serialization/send/channel-close/timeout), and by only inserting entries after successful request serialization.

This PR also adds a reusable leak-check workflow via `xtask`:
- `--probe lsp`: macOS `leaks` check for LSP open/change/close churn
- `--probe cli-timeout`: retained-memory growth probe for timeout/channel-close scenarios
- `--probe both`: runs both probes sequentially

### Leak root cause
`pending_requests` entries were inserted before serialization/send/wait finished and were not always removed when requests failed early. Repeated failing requests caused unbounded map growth and RSS growth.

### Why this fixes it
- serialize first, then insert into `pending_requests`
- on any async request error, remove the request id from `pending_requests` before returning

## Test plan
- Added pgls_cli regression tests:
  - `request_does_not_retain_pending_entries_when_serialization_fails`
  - `request_does_not_retain_pending_entries_on_timeout_or_channel_close`
- Ran:
  - `cargo test -p pgls_cli request_does_not_retain_pending_entries_when_serialization_fails`
  - `cargo test -p pgls_cli request_does_not_retain_pending_entries_on_timeout_or_channel_close`
  - `cargo run -p xtask -- leak-check --probe cli-timeout --iterations 200000`
  - `cargo run -p xtask -- leak-check --probe lsp --iterations 20000 --pause-ms 1`

### DIFF
--- a/xtask/src/leak_check.rs
+++ b/xtask/src/leak_check.rs
@@ -25,7 +25,9 @@ impl flags::LeakCheck {
         }
 
         if Command::new("leaks").arg("--help").output().is_err() {
-            bail!("`leaks` not found — install Xcode Command Line Tools (`xcode-select --install`)");
+            bail!(
+                "`leaks` not found — install Xcode Command Line Tools (`xcode-select --install`)"
+            );
         }
 
         let iterations = self.iterations.unwrap_or(DEFAULT_ITERATIONS);


xtask/src/leak_check.rs
```diff
@@ -0,0 +1,364 @@
...
+        .write_all(header.as_bytes())
+        .context("failed to write LSP header")?;
+    stdin
+        .write_all(payload.as_bytes())
+        .context("failed to write LSP payload")?;
+    stdin.flush().context("failed to flush LSP payload")
+}
+
+fn json_escape(s: &str) -> String {
+    let mut out = String::with_capacity(s.len());
```
juleswritescode: nit: wir pushen hier eh mehr chars
```suggestion
    let mut out = String::new();
```

xtask/src/leak_check.rs
```diff
@@ -0,0 +1,364 @@
...
+use crate::flags;
+
+const DEFAULT_ITERATIONS: usize = 200;
+const DEFAULT_PAUSE_MS: u64 = 20;
+
+impl flags::LeakCheck {
+    pub(crate) fn run(self, _sh: &Shell) -> anyhow::Result<()> {
+        if !cfg!(target_os = "macos") {
+            bail!("`xtask leak-check` is currently implemented only for macOS (`leaks` tool).");
+        }
```
juleswritescode, line 25: should we also check that `leaks` is installed? 

..we could say "please install xcode command line tools". Everybody loves to hear that 😂

xtask/src/leak_check.rs
```diff
@@ -0,0 +1,364 @@
...
+    stdin
+        .write_all(header.as_bytes())
+        .context("failed to write LSP header")?;
+    stdin
+        .write_all(payload.as_bytes())
+        .context("failed to write LSP payload")?;
+    stdin.flush().context("failed to flush LSP payload")
+}
+
+fn json_escape(s: &str) -> String {
```
juleswritescode: nit: we could use `serde_json::to_string()` instead

crates/pgls_cli/src/service/mod.rs
```diff
@@ -101,6 +101,11 @@ pub struct SocketTransport {
     pending_requests: PendingRequests,
 }
 
+#[cfg(not(test))]
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(15);
+#[cfg(test)]
+const REQUEST_TIMEOUT: Duration = Duration::from_millis(50);
```
juleswritescode: I think we should make this configurable — otherwise, long running integration tests would fail. Also, we could speed up the leak-script by setting duration to 2ms or so (no point in waiting for a dropped stream receiver)

xtask/src/leak_check.rs
```diff
@@ -297,15 +295,19 @@ fn run_lsp_churn(stdin: &mut ChildStdin, iterations: usize, pause: Duration) ->
...
         stdin,
-        r#"{"jsonrpc":"2.0","id":2,"method":"shutdown","params":null}"#,
+        serde_json::json!({"jsonrpc":"2.0","method":"exit","params":null}),
     )?;
-    send_lsp_message(stdin, r#"{"jsonrpc":"2.0","method":"exit","params":null}"#)?;
     Ok(())
 }
 
-fn send_lsp_message(stdin: &mut ChildStdin, payload: &str) -> anyhow::Result<()> {
+fn send_lsp_json(stdin: &mut ChildStdin, value: Value) -> anyhow::Result<()> {
```
juleswritescode, line 309: schick!

PR #698, opened 2026-02-27
fix(grammar): build grammar on CI

juleswritescode:
the 500.000+ lines just become too unwieldy — my local tools won't display diffs anymore because the git-diff is too large, github crashes, and so on.

Let's hope that CI generation doesn't take as long with caching.


crates/pgls_wasm/flake.nix
```diff
@@ -72,6 +73,7 @@
           nativeBuildInputs = with pkgs; [
             rustToolchain
             emscripten
+            tree-sitter
```
juleswritescode, line 76: not sure if we also need so specify a version for nix setup?

psteinroe: we can if this is important. it has a lockfile and we can override the version if needed

PR #701, opened 2026-03-01
test: verify all keywords of grammar are available in completions

juleswritescode:
Adds two tests that make sure:

1. new keywords added to `grammar.js` are included in the keyword-providers
2. all keywords in keyword-providers are included in `grammar.js`

I tried generating the tests from the `pg_query.proto` first, but some tokens such as "btree" and "gin" are no "official postgres keywords" — we do want to suggest them as keywords at the right spots, though.

Regarding keyword generation: I think generating the keywords from either protobuf or treesitter-grammar doesn't really make sense: We have the `.starts_statement()` and `.require_prefix()` builder pattern, and I figure there'll be more of those options soon; so we'd need some manual code anyways.


crates/pgls_treesitter_grammar/grammar.js
juleswritescode, line 1: these keywords were unreachable 👍🏻

juleswritescode:
> the KW list from postgres includes a keyword type to differentiate btw reserved and non-reserved keywords. did u check if the list we want here could be generated from that?

Yeah I did check it, but the issue is that we have some completable "keywords" that aren't even tokens in the Postgres protobuf. 

For example, for `create index my_cool_idx on auth.users using |`, we'd want to suggest `btree, hash, gist, spgist, gin, or brin`. These aren't even listed in the protobuf file.

So our "keyword completion" is actually more a "valid token at this position" completion.

PR #706, opened 2026-03-03
fix: prevent treesitter context panic on incomplete SQL

psteinroe:
Closes https://github.com/supabase-community/postgres-language-server/issues/704

**Root cause**: `ScopeTracker::register()` in `ancestors.rs` assumes the first node it encounters is always a scope boundary (`statement`, `ERROR`, or `program`). When tree-sitter parses malformed SQL, nodes like `function_arguments`, `:=`, `keyword_begin`, etc. can appear as the first traversed node, leaving `self.scopes` empty and hitting an `unwrap_or_else(panic!)`.

**Fix**:
- Changed the scope creation condition to also create a fallback scope when `self.scopes` is empty. This gracefully degrades (completions/hover return nothing useful instead of crashing).
- Added regression test that iterates all cursor positions in an incomplete function body to cover all node kinds from the issue.

### DIFF
--- a/crates/pgls_treesitter/src/context/ancestors.rs
+++ b/crates/pgls_treesitter/src/context/ancestors.rs
@@ -16,13 +16,13 @@ impl ScopeTracker {
     }
 
     pub fn register<'a>(&mut self, node: tree_sitter::Node<'a>, position: usize) {
-        if SCOPE_BOUNDARIES.contains(&node.kind()) {
+        if SCOPE_BOUNDARIES.contains(&node.kind()) || self.scopes.is_empty() {
             self.add_new_scope(node);
         }
 
         self.scopes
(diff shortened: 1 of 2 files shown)


juleswritescode:
Hmm, soooo the issue is actually that scope-boundaries are missing (block, transactions etc.).
The point of the panic was to alert us of missing boundaries, so I'd like to keep it there and maybe nudge the user to give us a good issue.

> etc. can appear as the first traversed node

that's not true — the issue is that a parent boundary was missing.


> to also create a fallback scope

is that outdated? I'm only seeing an `.is_empty()` check?

juleswritescode:
Added an alternative solution here: https://github.com/supabase-community/postgres-language-server/pull/709

PR #730, opened 2026-04-06
feat: expose new workspace function splitStatements with ranges

lneves12:

Exposes new workspace function splitStatements with ranges

## Additional context

It can be useful for custom postgres logic to expose this functionality, specially with ranges

### DIFF
--- a/crates/pgls_wasm/src/lib.rs
+++ b/crates/pgls_wasm/src/lib.rs
@@ -66,6 +66,18 @@ pub struct CompletionItem {
     pub detail: Option<String>,
 }
 
+/// A single SQL statement extracted from a multi-statement source.
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Statement {
+    /// The SQL text of the statement
+    pub sql: String,
+    /// Start byte offset in the original SQL string
+    pub start: u32,
+    /// End byte offset in the original SQL string
+    pub end: u32,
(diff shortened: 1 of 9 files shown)


crates/pgls_wasm/src/lib.rs
```diff
@@ -367,6 +379,24 @@ impl Workspace {
             Err(e) => vec![e.to_string()],
         }
     }
+
+    /// Split SQL into individual statements.
+    ///
+    /// Each returned `Statement` contains the SQL text and its byte offsets
+    /// in the original string.
+    pub fn split_statements(&self, sql: &str) -> Vec<Statement> {
+        let result = pgls_statement_splitter::split(sql);
```
psteinroe: what do you think of adding this as a method to the workspace and just calling into it here?

psteinroe (commented):
a small nit, rest lgtm! 

can you share more about the intended use case?

crates/pgls_wasm/src/lib.rs
lneves12: something like this ?

lneves12:
@psteinroe Thank you for your review!

> can you share more about the intended use case?

The use case is simple. I am building a SQL editor with support for postgres, so having this method exposed is quite useful to support running multiple queries at once or running only a set of queries (for instance the current query at cursor)

psteinroe (approved):
thanks for the contribution!

PR #733, opened 2026-04-23
Add is_incomplete to signal truncated completion

shihanng:

Modify completion response behavior to enable clients to properly handle cases where completion items exceed 50.

Prior to this change, the completion handler always returned `CompletionResponse::Array(items)`, which clients interpret per [LSP 3.17 spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/) as a complete list with `isIncomplete: false`. This caused clients to cache results and never re-request, even when the server truncated results at the 50-item LIMIT.

## What is the new behavior?

This change converts return paths to use [`CompletionResponse::List`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList)
with an explicit `is_incomplete` flag.

- On success, `is_incomplete` is set to true only when results were truncated at the LIMIT, signaling clients to re-query on the next keystroke for potentially more completions.
- On database connection errors, `is_incomplete` is set to true with an empty items list, signaling clients to retry once the connection recovers.

### DIFF
--- a/crates/pgls_lsp/src/handlers/completions.rs
+++ b/crates/pgls_lsp/src/handlers/completions.rs
@@ -29,7 +29,15 @@ pub fn get_completions(
         Ok(result) => result,
         Err(e) => match e {
             WorkspaceError::DatabaseConnectionError(_) => {
-                return Ok(lsp_types::CompletionResponse::Array(vec![]));
+                // The list is empty only because the DB is unreachable.
+                // Mark it incomplete so the client re-requests on the next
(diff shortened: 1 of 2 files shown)


psteinroe (approved):
thanks for the contribution!

shihanng:
Thank you for the review, @psteinroe. I see that some CI checks are failing, but I don't think they relate to the changes I made here. I've opened [another PR](https://github.com/supabase-community/postgres-language-server/pull/734) to address the CI issues.

PR #734, opened 2026-04-27
Update cargo-pgrx to 0.18.0 to match pglinter

shihanng:

Prior to this change, `cargo-pgrx` version 0.16.1 was specified in both the Dockerfile and GitHub Actions setup, while `pglinter`'s upstream repository had been updated to use `pgrx` 0.18.0.

See: <https://github.com/pmpetit/pglinter/blob/8be79efaacdc6b0a113b51b3f6dc8cf7cfe7680c/Cargo.toml#L19>

This change updates `cargo-pgrx` to version 0.18.0 in two places:

- `Dockerfile`
- `.github/actions/setup-postgres/action.yml`

We are seeing CI failures here:

- <https://github.com/supabase-community/postgres-language-server/actions/runs/24834883460/job/73077134950>
- <https://github.com/supabase-community/postgres-language-server/actions/runs/24834883460/job/73077134967>

## What is the new behavior?

The PR should fix the CI failures.

### DIFF
--- a/.github/actions/setup-postgres/action.yml
+++ b/.github/actions/setup-postgres/action.yml
@@ -69,7 +69,7 @@ runs:
         export PATH="$(pg_config --bindir):$PATH"
 
         # Install cargo-pgrx (version must match pglinter's pgrx dependency)
-        cargo install cargo-pgrx --version 0.16.1 --locked
+        cargo install cargo-pgrx --version 0.18.0 --locked
 
         # Ensure we build the extension for the host architecture (macOS-14 runners are arm64).
         # Release workflow also cross-compiles x86_64, but the local PostgreSQL installation is arm64.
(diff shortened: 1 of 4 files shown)


shihanng:
I'm seeing CI failures while working on [this PR](https://github.com/supabase-community/postgres-language-server/pull/733). I am bumping the version of `cargo-pgrx` to fix the issue, but I'm not sure if this is the right approach. 

Another option would be to pin the `pglinter` version to v1 🤔. Feel free to abandon this PR if this isn't the right way forward 🙏.

psteinroe (approved):
thanks for taking care!

shihanng:
It seems we're getting a [CI error](https://github.com/supabase-community/postgres-language-server/actions/runs/24978094714/job/73134049665?pr=734) due to a Rust toolchain version mismatch. [`pgrx` requires version 1.90.0](https://github.com/pgcentralfoundation/pgrx/blob/b2be3e1822b2e5769ba6253c828674d4c885ff01/rust-toolchain.toml#L2), but we have 1.88.0. If we can't bump the toolchain version, the least disruptive approach would be pinning the `pglinter` version to v1 🤔?

PR #736, opened 2026-05-18
fix: panic for plsql when followed of EOF

lneves12:

Handles EOF properly for plsql 

If you pass a string with a plsql statement at the end before EOF the wasm module will just panic and crash forever


lneves12:
@psteinroe let me know if this fix makes sense. We started seeing an error if you pass a statement like `\\something' to the splitter, the wasm module will just crash

psteinroe (approved):
thanks! 🫶

PR #746, opened 2026-05-25
fix(workspace): back off unreachable database diagnostics

psteinroe:
Prevent DB-backed workspace operations from retrying a known-unreachable database connection on every edit. Connection failures are cached per connection key in the connection manager with exponential backoff, and DB callers now route through the manager so failures are tracked consistently.

**Diagnostics Resilience**

Database-backed file diagnostics are best-effort. When the database is unreachable, parser and static diagnostics are still returned instead of failing the whole edit-triggered diagnostics request.

**Connection Manager Routing**

DB-backed diagnostics, schema-aware static analysis, splinter, completions, hover, statement execution, and schema-cache invalidation now use `ConnectionManager::with_pool`, so connection failure backoff is managed centrally.

**Regression Coverage**

Added a workspace test that points at a closed local Postgres port, verifies static diagnostics are still produced, and confirms the second diagnostics pull during backoff completes quickly.

### DIFF
--- a/crates/pgls_workspace/src/workspace/server.rs
+++ b/crates/pgls_workspace/src/workspace/server.rs
@@ -15,7 +15,7 @@ use document::{CursorPositionFilter, DefaultMapper, Document, FormatStatementMap
 #[cfg(feature = "db")]
 use document::{ExecuteStatementMapper, TypecheckDiagnosticsMapper};
 #[cfg(feature = "db")]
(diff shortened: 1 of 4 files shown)


juleswritescode (approved):
LGTM — one concern: we're still using `get_current_connection()` a whole bunch in `server.rs`, for example when we fetch diagnostics or completions. We should probably migrate that to `with_pool` as well

Issue #417, opened 2025-06-05
Grant Statements are not correctly split


juleswritescode:
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application.
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com).

## Describe the bug

Grant statements are not split properly.

<img width="777" alt="Image" src="https://github.com/user-attachments/assets/09727b9c-b1c8-4066-aff3-8137d107c21c" />

Synopsis from PostgreSQL docs:

```
GRANT { { CREATE | USAGE } [, ...] | ALL [ PRIVILEGES ] }
    ON SCHEMA schema_name [, ...]
    TO role_specification [, ...] [ WITH GRANT OPTION ]
    [ GRANTED BY role_specification ]

GRANT { { SELECT | INSERT | UPDATE | REFERENCES } ( column_name [, ...] )
    [, ...] | ALL [ PRIVILEGES ] ( column_name [, ...] ) }
    ON [ TABLE ] table_name [, ...]
    TO role_specification [, ...] [ WITH GRANT OPTION ]
    [ GRANTED BY role_specification ]
```

Issue #321, opened 2025-04-08
union (all) is not split properly


psteinroe:
`select 1 union all select 2` is split into two statements

psteinroe:
closed by @juleswritescode in #329

PR #759, opened 2026-06-10
fix(splitter): do not split EXPLAIN from the explained statement

amogiska:

Bug fix.

The statement splitter treats the statement after an `EXPLAIN` prefix as a new statement, so `EXPLAIN SELECT 1` is split into two statements (`EXPLAIN` + `SELECT 1`):

```js
splitStatements('EXPLAIN SELECT 1')
// => [{ sql: 'EXPLAIN' }, { sql: 'SELECT 1' }]
```

Downstream, `lint()` then reports valid SQL as invalid — the bare `EXPLAIN` fragment fails to parse with `Invalid statement: syntax error at end of input` — and consumers that execute the split statements send a broken bare `EXPLAIN` to Postgres. All EXPLAIN forms are affected (`EXPLAIN ANALYZE`, `EXPLAIN (ANALYZE, BUFFERS)`, etc.).

This is the same mis-split class previously fixed for UNION (#321) and GRANT (#417).

## What is the new behavior?

`EXPLAIN` is handled as a statement prefix: a dedicated `explain()` splitter function consumes the prefix — the parenthesized option list or the legacy `[ANALYZE | ANALYSE] [VERBOSE]` modifiers — and then delegates to the splitter function of the explained statement, mirroring how `statement()` dispatches. Covered forms include `EXPLAIN SELECT/INSERT/UPDATE/DELETE`, `EXPLAIN WITH ... SELECT`, `EXPLAIN CREATE TABLE AS`, `EXPLAIN VALUES`, and multi-statement sources (`EXPLAIN SELECT 1; SELECT 2` splits at the semicolon).

`EXPLAIN_KW` is deliberately not added to `STATEMENT_START_TOKENS`: `explain` is an unreserved keyword and remains valid as an identifier mid-statement (e.g. `SELECT explain FROM t`), which a break-on-EXPLAIN rule in `unknown()` would mis-split. A regression test covers this.

## Additional context

Found while integrating `@postgres-language-server/wasm` for SQL editing/execution at ClickHouse, where this surfaced both as lint false positives and as broken EXPLAIN execution.

### DIFF
--- a/crates/pgls_statement_splitter/src/lib.rs
+++ b/crates/pgls_statement_splitter/src/lib.rs
@@ -134,6 +134,39 @@ mod tests {
     }
 
     #[test]
+    fn explain_statements() {
+        let single_statements = vec![
+            "EXPLAIN SELECT 1;",
+            "explain select 1;",
+            "EXPLAIN ANALYZE VERBOSE SELECT * FROM contact;",
+            "EXPLAIN ANALYSE SELECT * FROM contact;",
+            "EXPLAIN (ANALYZE, BUFFERS, FORMAT JSON) UPDATE contact SET name = 'x';",
+            "EXPLAIN INSERT INTO contact (id) VALUES (1);",
+            "EXPLAIN DELETE FROM contact WHERE id = 1;",
+            "EXPLAIN WITH x AS (SELECT 1) SELECT * FROM x;",
+            "EXPLAIN CREATE TABLE contact_copy AS SELECT * FROM contact;",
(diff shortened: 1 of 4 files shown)


psteinroe:
thanks for the contribution! can you share more about your use case at clickhouse? happy to make the language server work for you!

psteinroe (approved):

lneves12:
@psteinroe hey! Clickhouse released recently a managed Postgres offering. We are using it inside our SQL Editor. Thank you for the great work so far!

Issue #744, opened 2026-05-24
Large file edits trigger too many sequential analysis requests (no coalescing/cancellation)


MegaCoolBlood:
# Description
When editing large SQL files, it appears that the full file content is sent to the language server on every change (potentially every keystroke), and each request is then processed one-by-one.

This causes a backlog:

If I make many quick edits, many requests are queued.
Responses arrive slowly and are effectively stale (for older document versions).
Even after I stop typing, the server continues processing old versions for a long time.
Example scenario

10 edit bursts
~10 keystrokes each
~100 analysis requests queued
I can finish editing in ~5 minutes, but diagnostics/completions keep catching up on outdated versions (response 15, 16, 18, etc.)
Expected behavior

Coalesce rapid document changes.
Cancel or skip outdated in-flight/queued analysis for old document versions.
Prioritize only the latest document snapshot once prior work completes.
Avoid processing stale versions when a newer version is available.
Suggested approach

Debounce didChange handling for expensive analysis.
Track document version and discard results for non-current versions.
Add cancellation tokens for long-running DB-backed analysis.
Limit queue depth per document to 1 pending “latest” request (replace older pending ones).
Why this matters
This creates noticeable lag and unnecessary load for both the language server and database, especially on large files and fast typing workflows.

psteinroe:
thanks for the feedback! will take a look what we can do.

MegaCoolBlood:
Thank you! I wanted to mention how much I love the project you did here and how committed you work on it!

mvanhorn:
Implemented in https://github.com/supabase-community/postgres-language-server/pull/761 - adds a 300 ms per-URL debounce in `Session` that cancels in-flight diagnostic tasks on each new `did_change`, so only one analysis run fires at the end of a keystroke burst.

PR #760, opened 2026-06-17
fix(lsp): discard stale diagnostics for outdated document versions

mvanhorn:
## Summary

Discard LSP diagnostics that were computed for an outdated document version, so stale results are never published over a newer (or freshly reopened) document.

`Session::update_diagnostics` awaits the DB-backed `pull_file_diagnostics`. During rapid edits to a large SQL file, each `textDocument/didChange` runs this synchronously, and responses for older document versions arrive after newer edits have already been applied. Previously those stale results were still published, causing diagnostics computed against superseded snapshots to overwrite newer ones.

This change captures the document version before computing diagnostics, re-reads the current version after the await point, and skips publishing when the versions no longer match. Any version mismatch is treated as stale (not just a strictly newer version), which also covers the close/reopen case where a client resets the document version to a lower number.

## Why this matters

Refs #744. The issue reports that editing large SQL files triggers a backlog of sequential analysis requests whose stale responses keep getting published after the user stops typing. Discarding results for non-current document versions is one of the fixes the reporter asked for ("Track document version and discard results for non-current versions"), and it removes the visible flicker of stale diagnostics with a minimal, low-risk change. The `Document` struct already carries `version`, so no schema change is needed.

This is a partial fix: debounce, in-flight cancellation tokens, and per-document queue-depth coalescing would reduce the redundant analysis work itself, but they require a dedicated server-side scheduler and are left as a separate, larger change. The version guard is independent and useful on its own.

## Testing

- Added a unit test for the version-comparison decision (`session::tests::stale_diagnostics_are_detected_by_version`), covering: a newer version arriving (stale), the same version (publish), and a reopened document reset to a lower version (stale). This test has no database dependency.
- `cargo check -p pgls_lsp`, `cargo fmt -p pgls_lsp --check`, and `cargo clippy -p pgls_lsp` all pass.
- The existing `pgls_lsp` integration tests are `#[sqlx::test]` and require a live Postgres test database; they were not run as part of this change (no behavioral change to those paths beyond the version guard).

### DIFF
--- a/crates/pgls_lsp/src/session.rs
+++ b/crates/pgls_lsp/src/session.rs
@@ -29,7 +29,7 @@ use tower_lsp::lsp_types::Url;
 use tower_lsp::lsp_types::{self, ClientCapabilities};
 use tower_lsp::lsp_types::{MessageType, Registration};
 use tower_lsp::lsp_types::{Unregistration, WorkspaceFolder};
-use tracing::{error, info};
+use tracing::{debug, error, info};
 
 pub(crate) struct ClientInformation {
     #[allow(dead_code)]
@@ -265,6 +265,7 @@ impl Session {
     pub(crate) async fn update_diagnostics(&self, url: lsp_types::Url) -> Result<(), LspError> {
         let pgls_path = self.file_path(&url)?;
         let doc = self.document(&url)?;
+        let captured_version = doc.version;
         if self.configuration_status().is_error() && !self.notified_broken_configuration() {
             self.set_notified_broken_configuration();
             self.client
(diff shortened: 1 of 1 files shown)


psteinroe (approved):
thanks for taking a look! I like the simplicity of it. I was planning to add a more sophisticated debounce layer but this is might even be sufficient already. Let's try!

mvanhorn:
Appreciate you merging this, @psteinroe. Discarding diagnostics computed for an outdated document version should stop the stale results.

mvanhorn:
Thanks @psteinroe - discarding diagnostics computed for an outdated document version should stop stale LSP results from clobbering newer ones.

psteinroe:
> Thanks @psteinroe - discarding diagnostics computed for an outdated document version should stop stale LSP results from clobbering newer ones.

thanks for taking care! pushing a new release now :)

PR #761, opened 2026-06-19
fix: debounce did_change diagnostics to prevent stale analysis backlog

### LINKED ISSUE #744: Large file edits trigger too many sequential analysis requests (no coalescing/cancellation)

mvanhorn:
## Summary

- Add a per-URL debounce map (`Arc<Mutex<FxHashMap<Url, AbortHandle>>>`) to `Session` that tracks one pending diagnostic task per open document.
- Replace the direct `session.update_diagnostics(url).await` call in `did_change` with `session.schedule_diagnostics(url)`, which cancels any in-flight task for that URL and spawns a new one that sleeps 300 ms before running analysis.
- Add `cancel_pending_diagnostics` called from `did_close` so no stale diagnostics fire after a document is closed.
- The existing `is_stale_diagnostics` version guard is retained as a second safety net for concurrent races.

## Why this matters

Fixes #744. When editing large SQL files, every keystroke sends a `textDocument/didChange` notification and the server was calling `update_diagnostics` synchronously for each one. This queued dozens to hundreds of back-to-back analysis requests (some involving DB-backed typechecking), wasting resources on both server and database. With this change, only the final change in a burst triggers an analysis run - matching editor conventions and preventing the stale-analysis backlog described in the issue.

## Testing

- Rapid-fire `did_change` notifications for the same URL: only one `update_diagnostics` call fires after the burst, not one per change (the abort path resets the timer each time).
- Two different URLs changing concurrently: each URL has its own debounce slot; neither blocks the other.
- A single isolated `did_change` with no follow-up: diagnostics still publish after the 300 ms delay.
- `did_close` while a debounce task is pending: `cancel_pending_diagnostics` aborts the task before the document is removed.
- Build: `cargo build -p pgls_lsp` and `cargo clippy -p pgls_lsp --all-targets` pass with no new errors or warnings.

### DIFF
--- a/crates/pgls_lsp/src/handlers/text_document.rs
+++ b/crates/pgls_lsp/src/handlers/text_document.rs
@@ -1,4 +1,4 @@
-use crate::{documents::Document, session::Session, utils::apply_document_changes};
+use crate::{documents::Document, session::SessionHandle, utils::apply_document_changes};
 use anyhow::Result;
 use pgls_workspace::workspace::{
     ChangeFileParams, CloseFileParams, GetFileContentParams, OpenFileParams,
@@ -9,7 +9,7 @@ use tracing::{error, field};
 /// Handler for `textDocument/didOpen` LSP notification
 #[tracing::instrument(level = "debug", skip(session), err)]
 pub(crate) async fn did_open(
-    session: &Session,
+    session: &SessionHandle,
     params: lsp_types::DidOpenTextDocumentParams,
 ) -> Result<()> {
     let url = params.text_document.uri;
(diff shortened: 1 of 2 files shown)


psteinroe (approved):
thanks for fixing this! also thanks for the comments :)

mvanhorn:
Thanks @psteinroe, follow-up landed - debouncing did_change diagnostics is a nice complement to the stale-version discard from the other PR.

Issue #692, opened 2026-02-25
False positive for pg_temp schema


mplanchard:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application.
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com).

This was asked as a question [here](https://github.com/orgs/supabase/discussions/12452), but it is a bug, since it rejects valid sql.

## Describe the bug

Each session has access to a temporary table schema, `pg_temp_nnn`, which is aliased to `pg_temp`. It is documented [here](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-SEARCH-PATH).

We often use `pg_temp` in our migrations to create temporary functions or tables that we do not need after the migration completes.

`postgrestools` reports an error for any use of `pg_temp`:

```
✖ schema "pg_temp" does not exist
```

However, this is not an error, as code using `pg_temp` runs just fine.

## To Reproduce

```sql
create table pg_temp.foo (id int);
select * from pg_temp.foo
```

## Expected behavior

The above should not yield an error.

## System information

- OS: linux

psteinroe:
thanks for reporting! Unfortunately, this is "by design" right now. Typecheck works by simply asking Postgres to `EXPLAIN` the query, and `pg_temp` does not exist there. We are now at a point where we might start with a custom type checker that does not rely on `EXPLAIN`. Can't promise it though. We could also add an intermediate fix that ignores `EXPLAIN` errors for tables and columns that are mutated in the same file. Will check.

PR #770, opened 2026-07-14
fix: typecheck false positive for SQL function args of table row type

### LINKED ISSUE #705: SQL Function: "missing FROM-clause" false positive when argument is row type
# Bug report

- [x] I confirm this is a bug with Supabase lsp, not with my own application.
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com).

## Describe the bug

postgres-language-server raises a false positive `missing FROM-clause entry for table "row_arg" (pg typecheck)` when accessing a parameter
 
- in an SQL function
- whose type is a composite type defined as a row of an existing table (see [relevant section in doc](https://www.postgresql.org/docs/current/xfunc-sql.html#XFUNC-SQL-COMPOSITE-FUNCTIONS) )

## To Reproduce

Here is a MRE:

```sql
CREATE TABLE public.tbl (
    id SERIAL PRIMARY KEY,
    name TEXT
  );

CREATE OR REPLACE FUNCTION public.get_tbl_name(row_arg public.tbl)
  RETURNS TEXT
  LANGUAGE sql
  STABLE
  AS $$
    SELECT row_arg.name;
  $$;
```

postgresql-language-server raises `missing FROM-clause entry for table "row_arg" (pg typecheck)`

## Expected behavior

There should be no error: this works at runtime because the argument is resolved.

## System information

- OS: macOS 26.1

## Additional context

This is similar to https://github.com/supabase-community/postgres-language-server/issues/353 , but I don’t think this is the same error, I think this case is specific to composite / row types. Solution might be to complete https://github.com/supabase-community/postgres-language-server/pull/366 to be able to mock those too.

mvanhorn:
## Summary

`pg typecheck` reported a false-positive `missing FROM-clause entry for table "row_arg"` when a SQL function argument's type is a table's row (composite) type, for example:

```sql
CREATE TABLE public.tbl (id SERIAL PRIMARY KEY, name TEXT);

CREATE FUNCTION public.get_tbl_name(row_arg public.tbl)
  RETURNS TEXT LANGUAGE sql AS $$
    SELECT row_arg.name;
  $$;
```

This is valid in Postgres because the argument resolves to the row, but the language server flagged `row_arg.name` as an error.

## Why this matters

Passing a table's row type as a function argument is a documented, common Postgres pattern. Reporting it as an error is a false positive: it flags correct code, which undermines trust in `pg typecheck` and can hide the real diagnostics a user cares about. This was reported in #705 with a minimal reproduction and confirmed by a maintainer.

## Cause

To type-check a SQL function body, `apply_identifiers` replaces `fn.arg[.field]` references with default literals, resolving the argument type through the schema cache's `types` list. The query that populates that list (`crates/pgls_schema_cache/src/queries/types.sql`) only admitted base/enum types and standalone composites created with `CREATE TYPE ... AS (...)` (`relkind = 'c'`). A table's implicit row type has `relkind = 'r'` (or `'p'` for a partitioned table), so it was never loaded. `resolve_type` then returned `None`, the `row_arg.name` reference was left unreplaced, and Postgres `prepare` reported the missing-FROM error.

## Fix

Extend `types.sql` to also load the composite row types exposed by relations (ordinary tables, partitioned tables, views, materialized views, and foreign tables) along with their column attributes, admitting `relkind IN ('c', 'r', 'p', 'v', 'm', 'f')` in both the attributes sub-select and the outer type filter. The attribute join is restricted to `attnum > 0` so relation system columns such as `ctid`/`xmin` are not mistaken for real fields. With the row type present in the cache, the existing composite-field-access branch in `resolve_type` mocks `row_arg.name` with the column's default literal, so the body type-checks cleanly. This completes the existing composite-type mocking path rather than depending on a separate type checker. The prepared `.sqlx` offline query cache is regenerated to match the updated query.

## Tests

- Integration (`crates/pgls_typecheck/tests/diagnostics.rs`): the reported MRE (`select row_arg.name` with `row_arg public.tbl`) now produces no diagnostic; a view-typed argument is covered as well.
- Unit (`crates/pgls_typecheck/src/typed_identifier.rs`): a table-backed row-type argument has its `.id` and `.name` field references replaced with the columns' default literals, and a system-column reference (`row_arg.ctid`) is deliberately left unresolved so it still surfaces downstream.

## Known limitations

Field access on row-type arguments uses the same identifier-substitution mocking as other function parameters, so it inherits that mechanism's existing limitations: a parameter name that also shadows a `FROM` alias, or an unqualified row-type name that is ambiguous across schemas under a custom search path, can still be substituted where Postgres would resolve differently. Addressing those cases belongs to the dedicated type checker discussed in #692 rather than to this query change.

Fixes #705

AI was used for assistance.


psteinroe (approved):
thanks for the contribution(s)!

mvanhorn:
Thanks for taking these, @psteinroe - table-row-type args no longer false-positive in typecheck.

Issue #769, opened 2026-07-12
Treesitter panic when procedure contains space after semicolon


gintsgints:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application.
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com).

## Describe the bug

Bug is repeatable with test in my fork - 
https://github.com/supabase-community/postgres-language-server/compare/main...gintsgints:postgres-language-server:func_with_leading_space_fails

## To Reproduce

Execute provided test

## Expected behavior

Server should not panic.

## System information

- OS: macOS

psteinroe:
thanks for the report! would you mind opening a pr? I see you already have a fix in the fork.

gintsgints:
Ok. I will do.

PR #773, opened 2026-07-22
Leading space fail

gintsgints:

Bug fix

Panic when procedure contains space after semicolon. Bug #769

## What is the new behavior?

No Panic :)


crates/pgls_treesitter/src/context/ancestors.rs
```diff
@@ -24,13 +24,17 @@ impl ScopeTracker {
     }
 
     pub fn register<'a>(&mut self, node: tree_sitter::Node<'a>, position: usize) {
-        if SCOPE_BOUNDARIES.contains(&node.kind()) {
+        // The `program` root node is never registered directly, so the first node
```
gintsgints, line 27: Not sure. Might be this comment is too much and should be removed.

psteinroe:
Thank you! @juleswritescode will take a look soon :)

juleswritescode:
Hey there! Just took a look at this — do you have a real world scenario where this leads to a bug? I can't reproduce it by typing `begin end; |` in a `test.sql` file and trying to get completions or hover results.

The reason I'm asking is that the treesitter context is best understood within the `statement-splitter` > `sanitization` > `treesitter-context` > `scope-tracker` chain.

e.g. the `sanitization` will modify the SQL before it's even passed to the context; the context will adjust the cursor-position to prevent out-of-statement issues, and so on.

Fixing it in the ScopeTracker seems like it's too low in the chain; other parts of the parser might still have trouble if there's a bug.

gintsgints:
To begin with, I have updated version of treesitter in my project - 0.26. Might be issue. I am investigating.

gintsgints:
Test in my fork branch without fix fails:

https://github.com/gintsgints/postgres-language-server/tree/func_with_leading_space_fails

Hope it is helpful.

juleswritescode:
Hey there!

Your branch fails for me with exactly 1 test, but it seems unrelated.

Could you give me an exact SQL file that breaks when I use the PGLSP on it?

Screenshot 2026-07-28 at 08 33 17

gintsgints:
omg. I forgot to push exact test.  (branch: `bump_treesitter_without_spacefix`)
I just had reduced my example to one in test - `BEGIN END; ` (space after ; is issue)

I can reproduce test with:
```
cargo test --package pgls_treesitter -- context::ancestors::tests::plpgsql_function_with_leading_space
```

PR #774, opened 2026-07-23
fix: parse support of delete() functions

lneves12:

Extensions like hstore or custom functions named delete break the parser as the parser thinks the token is a delete from statement

Extensions like hstore or custom functions named delete break the parser as the parser thinks the token is a delete from

## What is the new behavior?

Checks if the next character is a ( to distinguish both tokens


psteinroe (approved):

lneves12:
@psteinroe thanks for merging this one so quickly! When is the next npm release planned? Let me know if there is anything I can help with!

psteinroe:
> @psteinroe thanks for merging this one so quickly! When is the next npm release planned? Let me know if there is anything I can help with!

@lneves12 sorry, I created the draft release on friday but forgot to go through with it. it will be there any minute!

PR #777, opened 2026-07-25
fix(doc): update architecture description to match current code

gintsgints:

Docs update

## Additional context

Probably was very outdated. Not sure it is used. :)


psteinroe:
Thank you! These were indeed very outdated 😅

PR #783, opened 2026-09-07
Make query compatible with postgres 19

docteurklein:
postgres 19beta1 introduces a new column `location` in the `pg_available_extensions` view.  https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=f3c9e341c

But because its columns are referenced positionally, it breaks this query in 19.   Here is a fix that works on both 18 and 19.

Bug fix

  ✖ Database error: column reference "comment" is ambiguous
and
"cannot connect to database"


psteinroe (approved):

docteurklein:
ran and commited changes from `cargo sqlx prepare` to avoid the CI failure

Issue #780, opened 2026-08-16
Blank line after final CTE causes false syntax error


KvGroove:
# Blank line after final CTE causes false syntax error

* [x] I confirm this is a bug with Supabase, not with my own application.
* [x] I confirm I have searched the Docs, GitHub Discussions, and Discord.

## Describe the bug

A blank line between the closing parenthesis of the final CTE and the main statement causes `postgres-language-server` to report a syntax error for otherwise valid PostgreSQL.

This appears related to #372, which was closed by #380. However, the issue still reproduces on `postgres-language-server` 0.25.7 when the blank line occurs after the final CTE, before the main statement.

For example, this is reported as invalid:

```sql
WITH failed_ingestion_run AS (
    INSERT INTO ingestion_runs (
        source_name
    )
    VALUES (
        'test'
    )
    RETURNING id
)

INSERT INTO raw_fred_mortgage_rates (
    ingestion_run_id
)
SELECT id
FROM failed_ingestion_run;
```

The diagnostic is:

```text
Invalid statement: syntax error at end of input
```

Removing only the blank line makes the diagnostic disappear:

```sql
WITH failed_ingestion_run AS (
    INSERT INTO ingestion_runs (
        source_name
    )
    VALUES (
        'test'
    )
    RETURNING id
)
INSERT INTO raw_fred_mortgage_rates (
    ingestion_run_id
)
SELECT id
FROM failed_ingestion_run;
```

Both forms are valid PostgreSQL.

## To Reproduce

1. Use `postgres-language-server` 0.25.7.
2. Open a SQL file containing:

```sql
WITH foo AS (
    SELECT 1 AS id
)

SELECT id
FROM foo;
```

3. Observe the syntax diagnostic around the end of the CTE.
4. Delete the blank line so that `SELECT` immediately follows the closing `)`.
5. Observe that the diagnostic disappears.

The same behavior occurs when the main statement is an `INSERT`, for example with a data-modifying CTE using `INSERT ... RETURNING`.

## Expected behavior

Whitespace between the final CTE and its main statement should not change how the statement is parsed.

Both of these should be accepted without diagnostics:

```sql
WITH foo AS (
    SELECT 1 AS id
)

SELECT id
FROM foo;
```

and:

```sql
WITH foo AS (
    SELECT 1 AS id
)
SELECT id
FROM foo;
```

## Screenshots

N/A

## System information

* OS: Linux
* Editor: Neovim
[...]

Issue #372, opened 2025-04-24
Blank lines between common table expressions (CTEs) misreported as syntax error


aleclarson:
I assume this has to do with how statement splitting is implemented.

The SQL formatter I use likes to put blank lines between CTEs.

```sql
WITH users_cte AS (
  SELECT
    id,
    email
  FROM account
  WHERE tier = 'Pro'
),

active_sessions_cte AS (
  SELECT
    accountid,
    count(*) AS session_count
  FROM session
  WHERE expiresat > current_timestamp
  GROUP BY accountid
)

SELECT
  u.id,
  u.email,
  coalesce(s.session_count, 0) AS active_sessions
FROM users_cte AS u
LEFT JOIN active_sessions_cte AS s ON u.id = s.accountid
ORDER BY active_sessions DESC;
```

## Output

```
path/to/example.sql:7:3 syntax ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  × Expected Ident
  
     5 │   FROM account
     6 │   WHERE tier = 'Pro'
   > 7 │ ),
       │   
   > 8 │ 
   > 9 │ active_sessions_cte AS (
       │ 
    10 │   SELECT
    11 │     accountid,
  

path/to/example.sql:7:3 syntax ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  × Expected As
  
     5 │   FROM account
     6 │   WHERE tier = 'Pro'
   > 7 │ ),
       │   
   > 8 │ 
   > 9 │ active_sessions_cte AS (
       │ 
    10 │   SELECT
    11 │     accountid,
  

path/to/example.sql:7:3 syntax ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  × Expected Ascii40
  
     5 │   FROM account
     6 │   WHERE tier = 'Pro'
   > 7 │ ),
       │   
   > 8 │ 
   > 9 │ active_sessions_cte AS (
       │ 
    10 │   SELECT
    11 │     accountid,
  

Checked 1 file in 37ms. No fixes applied.
Found 3 errors.
[...]

PR #788, opened 2026-09-10
fix(splitter): preserve blank lines in parsable SQL

### LINKED ISSUE #784: Statement splitter breaks statements on blank lines, producing false syntax errors
## Describe the bug

The statement splitter treats a blank line as a statement boundary. Any blank line
*inside* a single statement causes that statement to be cut in two, and each half is
then reported as a syntax error by `libpg_query`.

Blank lines inside a statement are a common readability convention (separating join
blocks, separating a CTE list from the main query). On our codebase, this affects
**466 of 1436 SQL files** in one directory tree alone, which makes `check` unusable in
CI and very noisy in the editor.

## To Reproduce

`repro.sql`:

```sql
SELECT
	t.a
FROM t

LEFT JOIN u ON u.a = t.a;
```

```
$ postgres-language-server check repro.sql
repro.sql:5:1 syntax ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  × Invalid statement: syntax error at or near "LEFT"

    3 │ FROM t
    4 │
  > 5 │ LEFT JOIN u ON u.a = t.a;
      │ ^^^^^^^^^^^^^^^^^^^^^^^^^
```

Removing the blank line on line 4 makes the diagnostic disappear.

## Expected behavior

The statement is valid SQL and should parse cleanly. A blank line is not a statement
terminator in Postgres.

I understand the double-newline heuristic exists so the editor can recover from
incomplete statements while typing, which is the right tradeoff for the LSP. But it is
the wrong tradeoff for the CLI on committed files, where the input is expected to be
valid and complete.

Two options that would both work for us:

1. Only fall back to the blank-line heuristic when the statement fails to parse without
   it, rather than splitting eagerly.
2. Expose a configuration flag (or make it the default for the CLI, as opposed to the
   LSP) to split on `;` only.

## Related

This looks like the same family of bug as #372 / #380 (blank line between two CTEs) and
#780 (blank line after the final CTE, before the main statement) — both closed. This
report is a third trigger for the same underlying "blank line = statement boundary"
heuristic: a blank line inside a `FROM` clause, not adjacent to any CTE. Given three
separate triggers found so far, it might be worth addressing the heuristic itself (see
the two options above) rather than patching each shape individually.

## System information

- postgres-language-server 0.25.7
- macOS (aarch64), release binary `postgres-language-server_aarch64-apple-darwin`

edjubert:

Bug fix.

The statement splitter treats blank lines as statement boundaries. Valid SQL formatted
with blank lines can therefore be split into invalid fragments and reported as syntax
errors.

Fixes #784.

This also addresses the underlying behavior reported in #372, #380, and #780.

## What is the new behavior?

Statement splitting now uses two passes:

1. A coarse pass splits on `;` and structural statement starts while ignoring blank lines.
2. For fragments containing blank lines, a recovery pass checks whether PostgreSQL can
   parse the fragment:
   - parsable fragments remain intact;
   - unparsable fragments are re-split using blank lines as boundaries.

A blank line now splits SQL only when the complete fragment cannot be parsed.

This preserves recovery for incomplete SQL while typing, including cases such as a
missing `;` before `BEGIN`.

## Additional context

- Named parameters are normalized before the parser check, matching workspace parsing.
- The lexer records whether the source contains blank lines, so files without them avoid
  the recovery-path token scan.
- Valid fragments can still be split when their blank-line-separated parts are independently
  valid statements.

Added coverage for:

- blank lines inside a `FROM` clause;
- blank lines between CTEs and after the final CTE;
- blank lines inside string literals;
- named parameters;
- parser recovery for invalid SQL;
- a missing semicolon before `BEGIN`.

Validated with:

```text
cargo test -p pgls_lexer -p pgls_statement_splitter
cargo clippy --all-targets --all-features -- -D warnings

### DIFF
--- a/crates/pgls_lexer/src/lexed.rs
+++ b/crates/pgls_lexer/src/lexed.rs
@@ -30,6 +30,7 @@ pub struct Lexed<'a> {
     pub(crate) start: Vec<u32>,
     pub(crate) error: Vec<LexError>,
     pub(crate) line_ending_counts: Vec<usize>,
+    pub(crate) has_blank_line: bool,
 }
 
 impl Lexed<'_> {
@@ -69,6 +70,11 @@ impl Lexed<'_> {
         );
         assert!(self.kind(idx) == SyntaxKind::LINE_ENDING);
         self.line_ending_counts[idx]
+    }
+
+    /// Returns whether the source contains at least one blank line.
+    pub fn has_blank_line(&self) -> bool {
+        self.has_blank_line
     }
 
     /// Returns the text range of token at the given index
--- a/crates/pgls_lexer/src/lexer.rs
+++ b/crates/pgls_lexer/src/lexer.rs
@@ -12,6 +12,7 @@ pub struct Lexer<'a> {
     offset: usize,
     /// we store line ending counts outside of SyntaxKind because of the u16 represenation of SyntaxKind
     line_ending_counts: Vec<usize>,
+    has_blank_line: bool,
 }
 
 impl<'a> Lexer<'a> {
@@ -24,6 +25,7 @@ impl<'a> Lexer<'a> {
             error: Vec::new(),
             offset: 0,
             line_ending_counts: Vec::new(),
+            has_blank_line: false,
         }
     }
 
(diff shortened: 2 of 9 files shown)


psteinroe (approved):
thanks for pushing this!

PR #795, opened 2026-09-16
feat(format): add the commaStyle and logicalOperatorPlacement options

edjubert:
## What

Two new formatter options, both defaulting to the current behaviour:

| Option | Default | Values |
| --- | --- | --- |
| `commaStyle` | `"trailing"` | `"trailing"`, `"leading"` |
| `logicalOperatorPlacement` | `"trailing"` | `"trailing"`, `"leading"` |

They control where the separator sits **when a list or a condition breaks across lines**.
A list that fits on one line is unaffected.

### `commaStyle=leading`

```sql
select
	staging.buildings.identification_number
	, staging.buildings.construction_year
	, staging.buildings.address_fk
from
	staging.buildings;
```

### `logicalOperatorPlacement=leading`

```sql
where
	staging.buildings.construction_year > 1950
	and staging.buildings.address_fk is not null
	and staging.buildings.identification_number is not null;
```

with `trailing` (today's output, unchanged) the same condition ends each line with `and`.

## Why

Leading commas are a long-standing convention in analytics/ETL SQL: a line can be commented out or reordered without touching its neighbours, and a missing or extra comma is visible in a diff.
The same argument applies to `AND`/`OR` in long `WHERE` clauses.
Both are pure style preferences, hence options rather than a change of default.

## Implementation

The separator is chosen at emission time, not at render time: `node_list.rs` emits the comma before or after the line break depending on `commaStyle`, and `bool_expr.rs` does the same for the boolean operator.
That is why the emitter needs access to the configuration (see the prerequisite PR).

Wiring: `pgls_configuration` (serde + `bpaf` CLI flags `--comma-style` / `--logical-operator-placement`), `pgls_workspace` settings, `docs/features/formatting.md`, and `docs/schema.json`.

## Tests

Two new fixtures with snapshots at width 80, plus the full existing snapshot suite unchanged (the defaults preserve today's output).

## The two preparatory commits

The first two commits of this PR are shared, byte for byte, with the three other formatter-option PRs open alongside it:

* [#796 — `clauseBodyStyle` / `isolateSemicolon`](https://github.com/supabase-community/postgres-language-server/pull/796)
* [#797 — `layout`](https://github.com/supabase-community/postgres-language-server/pull/797)
* [#798 — `castStyle`](https://github.com/supabase-community/postgres-language-server/pull/798)

The shared commits are:

* `test(pretty-print): allow per-fixture format configuration`
* `refactor(pretty-print): thread the format config into the emitter`

They are carried in each PR so every option remains independently reviewable and mergeable.
Whichever PR lands first, I will rebase the other three to remove the duplicate commits.

Merge them in whatever order suits you; nothing here depends on the others.


psteinroe (approved):
nice!

PR #797, opened 2026-09-16
feat(format): add the expanded statement layout

edjubert:
## What

A new `layout` option:

| Option | Default | Values |
| --- | --- | --- |
| `layout` | `"fit"` | `"fit"`, `"expanded"` |

* `fit` - today's behaviour: a clause breaks only when the line would exceed `lineWidth`.
* `expanded` - `SELECT` target lists, joins and `INSERT` column lists always break, whatever the line width.

```sql
-- input
SELECT t.a, t.b FROM s.t WHERE t.c = 1 GROUP BY t.a ORDER BY t.a;

-- layout=expanded
select
	t.a,
	t.b
from
	s.t
where
	t.c = 1
group by t.a
order by t.a;
```

Joins each get their own line:

```sql
-- input: SELECT a FROM a LEFT JOIN b ON b.a = a.a JOIN c ON c.a = a.a;
select
	a
from
	a
	left outer join b on b.a = a.a
	inner join c on c.a = a.a;
```

`INSERT` with a CTE:

```sql
insert into t.x (
	a,
	b
)
with
c
as (
	select
		1 as a,
		2 as b
	from
		t.y
)
select
	c.a,
	c.b
from
	c;
```

## Why

With `fit`, whether a query is one line or ten depends on how long its identifiers happen to be.
Two queries with the same shape can be formatted differently, and adding a column can reflow an entire statement.
`expanded` makes the printed shape a function of the query structure alone, which keeps diffs minimal and reviewable - the property people want from a formatter in a versioned SQL repository.

## Implementation

A `clause_break(e)` helper in `nodes/mod.rs` emits a hard break under `expanded` and a soft one under `fit`.
`select_stmt.rs`, `join_expr.rs` and `insert_stmt.rs` call it where they previously emitted a soft line.
No new renderer state.

Wiring: `pgls_configuration` (`--layout`), `pgls_workspace` settings, `docs/features/formatting.md`, `docs/schema.json`.

## Tests

Three new fixtures snapshotted at widths 80 and 100 - identical at both widths, which is exactly the property `expanded` promises.

## The two preparatory commits

The first two commits of this PR are shared, byte for byte, with the three other formatter-option PRs open alongside it:

* [#795 — `commaStyle` / `logicalOperatorPlacement`](https://github.com/supabase-community/postgres-language-server/pull/795)
* [#796 — `clauseBodyStyle` / `isolateSemicolon`](https://github.com/supabase-community/postgres-language-server/pull/796)
* [#798 — `castStyle`](https://github.com/supabase-community/postgres-language-server/pull/798)

The shared commits are:

* `test(pretty-print): allow per-fixture format configuration`
* `refactor(pretty-print): thread the format config into the emitter`

They are carried in each PR so every option remains independently reviewable and mergeable.
Whichever PR lands first, I will rebase the other three to remove the duplicate commits.

Merge them in whatever order suits you; nothing here depends on the others.


psteinroe (approved):

edjubert:
Restacked this PR on #795 so expanded lists can honor `commaStyle`. The added regression test covers both `INSERT` target columns and `SELECT` target lists with `layout=expanded` and `commaStyle=leading`.

PR #803, opened 2026-09-16
fix(pretty-print): suppress child statement semicolons in CREATE SCHEMA

edjubert:
## What

`CREATE SCHEMA` takes a list of schema elements.
Per the grammar (`OptSchemaEltList: OptSchemaEltList schema_stmt`) they are space-separated and only the outer statement carries a terminator.
The emitter printed each child statement's own `;`, producing invalid SQL:

```sql
-- input
CREATE SCHEMA s1 CREATE TABLE a (f1 int) CREATE TABLE b (f2 int);

-- before: does not parse
create schema s1 create table a (f1 int);create table b (f2 int);;

-- after
create schema s1 create table a (f1 int) create table b (f2 int);
```

## Implementation

`EventEmitter` gains a `no_semicolon` flag that makes `token()` drop `SEMICOLON`; it is set while emitting `schema_elts` and cleared afterwards.
The separator before the element list becomes a soft line so the renderer can wrap when the schema name and the first element do not fit.

## Tests

`tests/create_schema_elts.rs` round-trips a `CREATE SCHEMA` with two elements: format, reparse, normalize, compare ASTs.


psteinroe (commented):
thanks for all the work and sorry for the late review!

I appreciate the attempt to keep the total loc changes low but would like to avoid adding sql context state to the event emitter. can we follow the existing `*_no_semicolon` pattern instead? add no-semicolon variants for the six supported schema-element statement types and dispatch these from `CREATE SCHEMA`. 

we could also explore moving terminator ownership to statement boundaries but that would be a broader refactor and not needed for this edge-case fix.

